PR #1, opened 2024-05-24
Add excludeMain() functionality in AST check

sarpsahinalp:
### Motivation and Context

Improved the current AST functionality to allow instructors to exclude the Main method from AST checks. This change ensures that when students use for loops or recursion within the Main method for testing purposes, their tests will still pass.

### Description

- Added a method, `excludeMain()`, to avoid AST checking of the Main method by removing it from the AST.
- The functionality is controlled via a `boolean` flag which is set to false by default.
- The boolean flag is passed down to `JavaFile` and removes the node associated with the `Main` method from the created `CompilationUnit` (AST).

### Steps for Testing

Prerequisites:
- 1 Programming Exercise with Ares

1. Run the `AstAssertionTests` and ensure that they pass.
2. Create an example exercise template with a main method, but include an `Unwanted type` that you want to detect.
3. Test with `UnwantedNodesAssert` that the method fails with an error message similar to the one in the screenshot below.
4. Now, include the `excludeMain()` method, and the test will pass since the main method won’t be considered as an Unwanted type.
5. Here’s an example test:
```java
@Test
void testHasBelowNoLoopsOutsideMainMethod_Success() {
	UnwantedNodesAssert.assertThatProjectSources().withinPackage(BASE_PACKAGE + ".excludeMain.no")
		        .excludeMainMethod()
                        .withLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_17)
			.hasNo(LoopType.ANY);
}
```

### Screenshots

image

### DIFF
--- a/src/main/java/de/tum/cit/ase/ares/api/ast/asserting/UnwantedNodesAssert.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/ast/asserting/UnwantedNodesAssert.java
@@ -34,8 +34,18 @@ public class UnwantedNodesAssert extends AbstractAssert<UnwantedNodesAssert, Pat
 	 */
 	private final LanguageLevel level;
 
+	/**
+	 * Whether to exclude the main method for the Java parser Default is set to true
+	 */
+	private final boolean excludeMainMethod;
+
 	private UnwantedNodesAssert(Path path, LanguageLevel level) {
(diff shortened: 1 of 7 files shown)


MarkusPaulsen (approved):
Code looks good + Tested on Sarp's machine.

PR #4, opened 2024-07-03
`ArchUnit`: Filesystem access extension for Ares

sarpsahinalp:
Description:
- Analyzed native methods used to access the filesystem
- Created the initial Architecture for the `ArchUnit` security tests extension
- Implemented an efficient algorithm (`ArchCondition`) to parse thousands of .class files in bytecode and check for their transitive Path of method calls in `TransitivelyAccessesMethodsCondition`!
- Created the First implementation for the SecurityRule `noClassesShouldAccessFileSystem`

Problems:
- Since we are analyzing bytecode mainly top level interfaces, abstract classes are registered as a method call therefore the implementation of some methods are not present (needs to be resolved during runtime) &#8594; this is already enough for banning FileAccess completely with the examples I have tested

Steps for Testing: 
- Try out all different ways of accessing the filesystem, take a look at the Java 21 Docs for file access methods
- (?) Test with student submission from an exercise that needed filesystem access
- Analyze which top level methods calls the securiy violating methods


src/main/java/de/tum/cit/ase/ares/api/archunit/SecurityRuleExecutor.java
```diff
@@ -0,0 +1,97 @@
...
+                .map(d -> d.getTargetClass().getPackageName())
+                .collect(Collectors.toSet());
+    }
+
+
+    /**
+     * Executes the security rules
+     */
+    public void executeSecurityRules() {
+        Set<String> packages = getDependencies(classFileImporter.importPackages(studentSubmissionPackage));
```
Strohgelaender, line 80: The selection of the correct packages is something we need to be quite careful about. Students could create their own `org.junit` package and implement their own versions of some classes there, that then will get prefered by the classloader.

Ares1 solved this by having a well-defined set of trusted packages and verifying that these packages are not present in the student submission. Here we would need to go the other way around since a student could just create a `mypackage` package that would then not be analyzed by ArchUnit.

sarpsahinalp: ArchUnit works with JavaBytecode and looks at the classes in target/classes and other packages, we can therefore also here check for which packages student have created within the submission by getting their packages, the question would be what happens if they change it and the code doesn't compile on our side (in the test repository) or changes them somehow to not execute the test but pass immediately. I don't know what would happen then. We would have to check this before the exeuction of the tests.

sarpsahinalp: I agree thanks for your input!

src/main/java/de/tum/cit/ase/ares/api/archunit/SecurityRules.java
```diff
@@ -0,0 +1,56 @@
...
+                "java.util.jar",
+                "sun.awt.X11",
+                "javax.imageio.stream",
+                "javax.sound.midi",
+                "javax.swing.filechooser",
+                "java.awt.desktop");
+
+        return ArchRuleDefinition.noClasses()
+                .that()
+                .resideInAPackage(studentPackage + "..")
```
Strohgelaender, line 43: same issue here, none of the analyzed classes are allowed to access the file system.
```suggestion
```

PR #44, opened 2025-03-10
Feature/aspectj instrumentation for thread creation

MarkusPaulsen:


src/test/java/de/tum/cit/ase/ares/integration/precompile/PrecompileTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+                });
+            }
+        } catch (IOException e) {
+            System.err.println("Error cleaning directory: " + e.getMessage());
+        }
+    }*/
+
+    @Test
+    void testPrecompileJavaMavenArchunitInstrumentation() {
```
AjayvirS, line 40: remove hardcoded data with empty strings and add todo

src/test/resources/tempMethods.txt
AjayvirS, line 1: do we need this file?

MarkusPaulsen: I would keep it in for now.

Predixx (approved):
Added tests not fully working yet but can be fixed later

PR #56, opened 2025-06-15
`Ares`: Fixed architecture tests.

MarkusPaulsen:
This pull request fixes the architecture tests by introducing a two step test procedure:
- The critical resource is accessed via a deactivated test that is annotated with a policy file
- The deactivated test is called by a separate test, which can catch and check the thrown SecurityException. The deactivated test would not be able to do that, as the SecurityException is thrown before the start of the test case (which is a security feature).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test coverage for file system access restrictions, ensuring that disabled tests are now programmatically executed and verified for expected security exceptions.
  - Corrected import paths for test classes to ensure proper test execution.

- **Chores**
  - Removed numerous obsolete test utility classes and methods related to file and thread system access.
  - Updated security policy configurations to explicitly list relevant test classes.
  - Cleaned up unused resource and text files.


Predixx (approved):
LGTM!

PR #61, opened 2025-06-27
Chore/fix file overwrite delete test cases

AjayvirS:
## Summary
Extend the security test matrix so that all relevant delete APIs are covered and remove false-positive failures caused by helper threads that some JDK classes spawn.:contentReference[oaicite:0]{index=0}

## Changes
- Added `wrapIgnoringJdkHelpers` decorator and supporting constants in `JavaWalaTestCaseCollection.java`; WALA now ignores call chains that consist entirely of trusted JDK housekeeping classes.:contentReference[oaicite:1]{index=1}  
- Registered ten additional delete entry-points (File, Files, SecureDirectoryStream, `DELETE_ON_CLOSE`, etc.) with dedicated subject mains and assertions in `FileSystemAccessDeleteTest`.:contentReference[oaicite:2]{index=2}  
- Moved all delete fixtures into `src/test/.../fileSystem/delete/**` and updated four policy YAML files so that test code no longer lies inside the protected path.:contentReference[oaicite:3]{index=3}  
- Replaced generic `assertAresSecurityExceptionRead` with `assertAresSecurityExceptionDelete` and deleted the redundant Path→File AspectJ test.:contentReference[oaicite:4]{index=4}  
- Minor clean-up: added toolbox TODO in `JavaInstrumentationAgent.java` and removed an unused FileSystemProvider method from the point-cut list.:contentReference[oaicite:5]{index=5}

## Rationale
Before this change, only a subset of delete operations was checked and tests failed whenever the JDK opened internal selector or poller threads. The new ignore logic keeps the security guarantee intact while preventing these false positives, and the broader test set blocks additional file-system shortcuts.:contentReference[oaicite:6]{index=6}

## Testing
```bash
mvn clean verify -Pintegration-tests


MarkusPaulsen (approved):
Code looks good

Predixx (approved):

PR #97, opened 2026-05-28
fix(aop): correct allowlist path matching for non-existing policy entries

SedaOran:
## Summary

- **Bug 1 (allowlist silently dropped):** `variableToPath` was called with `allowNonExistingPathsToBeConsidered=false` when resolving *policy* entries. For read-operations this caused any allowlist rule whose file did not yet exist on disk to be silently skipped, making the entire allowlist unreachable.
- **Bug 2 (pathMatches wrong):** `pathMatches` returned `false` (no match) whenever the *allowed path from the policy* did not exist, regardless of whether the candidate path matched lexically.
- Both bugs affected `JavaInstrumentationAdviceFileSystemToolbox` (Byte Buddy mode) and `JavaAspectJFileSystemAdviceDefinitions` (AspectJ mode).

**Effect:** Explicitly allowed resources were still blocked (SecurityException on allowlisted files), and wrong-target policy tests did not throw SecurityException because access to non-existing files bypassed the allowlist check entirely.

## Related Issues
Fixes: Allowed Resource Access Fails (Issues #6, #7) and WrongTarget Policy Violation in AspectJ Modes (Issues #2, #3).

## Test plan
- [ ] Define a policy allowing `protected/allowed.txt` for reads
- [ ] Run `Files.readString(Path.of("protected/allowed.txt"))` — should PASS
- [ ] Run `Files.readString(Path.of("protected/other.txt"))` — should throw SecurityException
- [ ] Run `WrongTargetPolicyMatrixTest` in ARCHUNIT+ASPECTJ and WALA+ASPECTJ Gradle modes — all `wrongTarget_*` assertions should pass

🤖 Generated with [Claude Code](https://claude.ai/code)

## Bug Fixes
* Improved filesystem allowlisting so permitted paths are correctly matched even before the target exists.
* Refined execution security checks to detect and report more violations consistently.
* Made architecture method-signature matching more robust, reducing incorrect classifications.

## New Features
* Added configuration-driven allowlisting for executable file paths to control which paths may be executed.

## Tests
* Updated security and integration tests to reflect the revised execution/exception behavior and adjusted messaging.


MarkusPaulsen (approved):
Code looks good

ShudongCai (approved):
LGTM

PR #86, opened 2026-03-23
Bugfix: Enhance security policy handling and improve test file retrieval

MarkusPaulsen:
## Summary

Fixes security enforcement gaps and extends Ares with JNDI-injection detection, improved path/command matching, and comprehensive timeout policies — ensuring student code is always restricted unless explicitly opted out.

### Security enforcement (default-on)
- **`Policy.java`**: Updated Javadoc to clarify that `activated = true` (default) enforces Ares security; `false` explicitly opts out.
- **`JupiterSecurityExtension.java`**: Flipped `isPolicyActivated` default from `false` to `true`. Security settings are now unconditionally reset before and after every test. Extracted `resetSettingsInStandardClassLoader()` as a reusable public helper.
- **`JqwikSecurityExtension.java`**: Aligned with Jupiter logic — always resets AOP settings before/after each test; enforces security by default; only skips enforcement when `@Policy(activated=false)` is explicitly set.
- **`SecurityPolicyReaderAndDirector.java`**: Restored the fallback that applies the most-restricted default policy when no policy file is specified (`.or(...)` instead of `.orNull()`).

### JNDI-injection architecture check
- **`JavaArchitectureTestCaseSupported.java` / `JavaArchunitTestCase.java` / `JavaArchunitTestCaseCollection.java`**: Added `JNDI_INJECTION` as a new supported architecture test case.
- **`jndi-injection-methods.txt` / `JNDI_INJECTION.txt`**: New resource files listing JNDI-related methods and the corresponding ArchUnit rule template.
- **`JavaArchitectureTestCase.java`**: Wired JNDI check into architecture test dispatch.

### Path & command matching improvements
- **`JavaAspectJFileSystemAdviceDefinitions.aj` / `JavaAspectJFileSystemPointcutDefinitions.aj`**: Refined file-system pointcuts and advice for more accurate path matching.
- **`JavaInstrumentationAdviceFileSystemToolbox.java` / `JavaInstrumentationAdviceAbstractToolbox.java`**: Improved path normalisation and call-stack inspection logic.
- **`JavaInstrumentationExecuteCommandMethodAdvice.java` / `JavaInstrumentationPointcutDefinitions.java`**: Tightened command-execution pointcut matching.
- Expanded `file-system-access-methods.txt`, `command-execution-methods.txt`, `network-access-methods.txt`, and `thread-manipulation-methods.txt` (both ArchUnit and WALA variants) with additional intercepted methods.

### Timeout policies
[...]

PR #99, opened 2026-05-28
fix(security): do not enforce security when no @Policy annotation is present

SedaOran:
## Summary

- **Root cause:** When a test has no `@Policy` annotation, `scanForPackageName()` returns `""` (empty string) as the restricted package (the default when no package is found). Since every class name starts with `""`, the AOP advice treated every stack frame as restricted, causing `SecurityException` in trivial tests like `@PublicTest void simpleTest() { assertTrue(true); }`.
- **Fix:** In `JupiterSecurityExtension`, security enforcement (`createTestCases` + `executeTestCases`) is now only activated when an explicit `@Policy` annotation is present. Without `@Policy`, settings are reset to a clean state but no restrictions are applied.
- Also removes leftover `System.err` debug logging that was left in the `beforeTestExecution` callback.

## Related Issues
Fixes: Minimal `@PublicTest` / `@HiddenTest` Fails — Issue #9 from the security audit.

## Test plan
- [ ] `@PublicTest void simpleTest() { assertTrue(true); }` — should PASS without any `@Policy`
- [ ] `@HiddenTest void hiddenSimple() { assertTrue(true); }` — should PASS without any `@Policy`
- [ ] Tests WITH `@Policy` still enforce the security rules correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/main/java/de/tum/cit/ase/ares/api/jupiter/JupiterSecurityExtension.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/jupiter/JupiterSecurityExtension.java
@@ -36,7 +36,6 @@ public final class JupiterSecurityExtension
 	 */
 	@Override
 	public void beforeTestExecution(ExtensionContext extensionContext) throws Exception {
-		System.err.println("[ARES-LIFECYCLE] beforeTestExecution called");
 		resetSettingsInStandardClassLoader();
 		resetSettingsInBootstrapClassLoader();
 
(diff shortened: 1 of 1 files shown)


ShudongCai (commented):
This PR seems to conflict with a PR that has already been merged: https://github.com/ls1intum/Ares2/pull/86

As far as I understand, the intended behavior introduced there was that the security mechanism should still be enforced even when no `@Policy` annotation is present.

PR #100, opened 2026-05-28
feat(aop): add denial reason to file-system SecurityException messages

SedaOran:
## Summary

- All AOP `SecurityException`s previously had the same generic form with no indication of *why* the access was blocked, making policy development difficult. This PR adds the missing denial reason across all four runtime subsystems (file system, network, command, thread), in both the Byte Buddy instrumentation and AspectJ back-ends.
- Two new message keys introduced:
  - `security.advice.denial.reason.no.allowlist` → "No allow rule configured for this resource type."
  - `security.advice.denial.reason.not.in.allowlist` → "No configured allow rule permits this access."
- A `buildDenialReason(boolean noAllowRuleConfigured)` helper added to both `JavaInstrumentationAdviceAbstractToolbox` (Byte Buddy) and `JavaAspectJAbstractAdviceDefinitions` (AspectJ).
- Every SecurityException throw-site now appends the reason after a `|` separator — file system (3), command (2), network (4) and thread (2) per back-end. Each subsystem derives its own "no allow rule configured" flag from its allow-lists.

**Example output after fix:**
```
Ares Security Error: com.example.Main.readFile tried to illegally read File /secret.txt
via Files.readString | Reason: No allow rule configured for this resource type.
```

## Related Issues
Fixes: Generic / Non-informative Security Exceptions — Issue #5 from the security audit.

## Test plan
- [x] With a policy allowing only one read path, read a non-allowed file and verify the exception message ends with "No configured allow rule permits this access." (covered end-to-end for file, command and thread across all four ArchUnit/WALA × AspectJ/Instrumentation runners, and for network on the AspectJ runners).
- [x] Invoke each subsystem's advice toolbox with an empty allow-list and verify the message ends with "No allow rule configured for this resource type."
- [x] Invoke each subsystem's advice toolbox with a partial allow-list and verify the message ends with "No configured allow rule permits this access."

_Note: the "no allow rule configured" reason is only asserted at the toolbox level — with an empty policy the static ArchUnit/WALA layer blocks the capability before the runtime AOP layer adds the reason. Network is end-to-end tested under AspectJ only, since the instrumentation path does not intercept `java.net.Socket` connections._

## Summary by CodeRabbit

* **New Features**
  * Improved security violation error messages with clearer denial reasons, distinguishing between "no allow rule configured" and "access not permitted by configured rules."

* **Bug Fixes**
  * Enhanced error handling to gracefully skip field access during security checks when reflection fails, improving robustness of security checks.

* **Removals**
  * Removed architecture testing framework (ArchUnit-based test generation and validation system).


ShudongCai (commented):
Thanks for the PR! I agree that this is helpful for debugging, but I do not think we should expose the detailed denial reason directly in the `SecurityException` message, since students can also see this output.

This could leak information about the policy structure, e.g. whether a resource type has no allow rule at all or whether only the specific path is outside the allowlist. I would prefer to keep the student-facing message generic and only include the detailed reason in internal/debug logs or behind a configuration flag.

ShudongCai (approved):
Code LGTM

ShudongCai (approved):
reapprove

MarkusPaulsen (approved):
Code looks good

PR #126, opened 2026-07-18
Harden policy and enforcement for Ares 2 Beta9

MarkusPaulsen:
## Overview

This updates Ares 2 to **2.0.1-Beta9** and implements all 26 policy and enforcement audit actions. The central change is that the YAML security policy is now the single, reliable source of truth: Beta9 detects projects more accurately, rejects malformed policies earlier, and closes lifecycle and runtime-enforcement gaps.

### 1. Supervised tests behave more predictably

- Method-level `@Policy` annotations take precedence over class-level policies.
- Class-level policies apply consistently to nested test classes.
- Without a policy, Ares automatically detects Maven or Gradle and applies the restrictive ArchUnit + AspectJ default.
- Security preparation happens exactly once per test invocation.
- Security state is reliably cleared afterwards, including when preparation or execution fails.
- Constructors and lifecycle methods are no longer mistaken for test executions.
- Deadline handling is shared between JUnit and jqwik. Method deadlines override class deadlines, extensions are applied consistently, and public tests cannot evade the deadline prohibition through an inherited class deadline.

### 2. Policy files now fail closed

Beta9 validates the complete YAML structure before constructing a policy. It rejects:

- missing or null required values;
- incorrect scalar types, including numbers used as booleans;
- non-integer ports, thread counts or timeouts;
- unknown or misspelt fields;
- duplicate YAML keys;
- multiple YAML documents in one file;
- a null YAML root;
- malformed command permissions; and
- blank class, command, package or host names.

Command mappings must contain exactly `executeTheCommand` and `withTheseArguments`. A bare command string remains supported and means “allow this command without arguments”. Policy filename extensions are case-insensitive, including `.yaml`, `.YAML`, `.yMl` and `.YML`.

### 3. Placeholder expansion is safer

Placeholders such as `${PROJECT_ROOT}` are expanded only after YAML parsing. Replacement values containing colons, hashes, quotes or other YAML-sensitive characters therefore cannot alter the document structure. `${PROJECT_ROOT}` now refers to the explicit Ares project rather than the process working directory.

### 4. Network permissions have one wildcard

Port `0` is consistently the “any port” wildcard across policy validation, AspectJ and instrumentation. Ports `1..65535` select one port; negative values, including the previous inconsistent `-1` interpretation, are rejected.

### 5. Maven and Gradle discovery is stronger

Ares recognises `pom.xml`, `build.gradle` and `build.gradle.kts`, along with Maven custom source directories, common Groovy and Kotlin Gradle source-set declarations, Gradle-property paths, and multiple source roots.

If both Maven and Gradle descriptors exist without an explicit selection, Ares reports the ambiguity rather than guessing. Unsupported projects also fail clearly, and every discovered source root must remain inside the selected project.

### 6. Java source scanning understands Java

The regular-expression scanner is replaced with JavaParser syntax-tree analysis. It distinguishes code from comments and strings and recognises classes, records, interfaces, enums, nested and package-private types, semantic main methods, JUnit 4, Jupiter and jqwik tests, parameterised and dynamic tests, and meta-annotations.

Malformed or unreadable source files are reported with their paths, and package tie-breaking is deterministic.

### 7. Generated files cannot escape the project

The Ares writer receives an explicit authorised project root. Every output path is canonicalised and checked before writing, preventing malformed package or target paths from redirecting generated security files outside the exercise.

### 8. Runtime enforcement is more robust

- Thread creation is recorded before a short-lived thread can disappear, closing a race condition.
- AspectJ avoids duplicate checks around its own call-site wrappers.
- Essential packages are included consistently in generated allowlists.
- Runtime settings are published only after all domain allowlists are initialised.
- Network, filesystem, command and thread behaviour is aligned across AspectJ, instrumentation and Phobos.

The static/runtime boundary is now documented explicitly: static analysis denies an entire domain while it has no allowance; once one permission exists, runtime enforcement distinguishes that permission from every other operation.

### 9. Reserved packages have a versioned build boundary

Beta9 ships versioned Maven and Gradle build checks for packages that student code must not impersonate, including Java, Ares, AspectJ, WALA, ArchUnit and Byte Buddy infrastructure. There is deliberately no supported skip property because disabling this check removes part of the security boundary.

### 10. The unused legacy annotation system is removed

The old whitelist and blacklist annotations produced a detached configuration that the active enforcement pipeline never consumed. They have been removed so the API no longer suggests protection that does not exist.

This includes path, package and class whitelist/blacklist annotations, `@AddTrustedPackage`, `@AllowLocalPort`, `@AllowThreads`, `@TrustedThreads`, `AresSecurityConfiguration`, and its builder. `@Policy` with YAML is now the sole authority. The new enforcement-model document contains the complete migration table.

### 11. Source-level migration points

- `SupervisedCode.theFollowingClassesAreTestClasses` now uses an immutable `List<String>` rather than `String[]`.
- `CommandPermission.createRestrictive(command)` becomes `allowWithoutArguments(command)`.
- `PackagePermission.createRestrictive(packageName)` becomes `allowPackage(packageName)`.
- `@StudentCompiledClassesPath` may only be placed on a class.

## Root cause

The audit found inconsistent policy parsing, project discovery based on narrow textual assumptions, duplicate test-extension preparation, detached legacy configuration APIs, and runtime edge cases whose behaviour differed between enforcement backends. Several accepted design constraints also lacked an explicit deployment contract. This change consolidates those paths, makes malformed or ambiguous input fail closed, and documents the remaining trust boundaries.

## Upgrade checklist

1. Update the dependency to `2.0.1-Beta9`.
2. Replace removed security annotations with `@Policy` and YAML entries.
3. Ensure every required YAML field is present with the correct type.
4. Change an old any-port value of `-1` to `0`.
5. Replace programmatic test-class arrays with `List.of(...)`.
6. Integrate and retain the supplied reserved-package build check.
7. Keep `withinPath` under instructor control; it remains trusted configuration.

## Validation

- `mvn install` — **BUILD SUCCESS**
- 2,386 tests, 0 failures, 0 errors
- Checkstyle, PMD, CPD and SpotBugs pass
- Beta9 main, agent, sources and Javadoc artefacts install successfully
- `git diff --check` passes


ShudongCai (approved):
Code LGTM

PR #137, opened 2026-07-21
Make PACKAGE_IMPORT able to forbid individual java.* packages

MarkusPaulsen:
## Summary

Make `PACKAGE_IMPORT` able to forbid individual `java.*` packages. The essential allow-list previously whitelisted the entire JDK through a single subtree-matched `java` entry, so no `java.*` package could be forbidden; this adds exact-match semantics and narrows the baseline to `java.lang`.

## Linked issues

None.

## 1. Problem

Observed under both AOP modes and both architecture modes (WALA does not use its call graph for this category; it delegates to the same ArchUnit rule). `PACKAGE_IMPORT` could not forbid any `java.*` package.

The allow-list is assembled in `JavaCreator.prepareAllowedPackages` from `EssentialPackages.yaml`, which declared the bare root `java`. Matching was subpackage-inclusive (`startsWith`), so that one entry whitelisted the entire JDK. Reproduced against the real rule, with a class importing `java.io.File`, `java.net.Socket` and `java.lang.reflect.Method`:

```
allowed = [java, student]      (shipped config)  -> PASSED, nothing blocked
allowed = [java.lang, student]                   -> java.io + java.net blocked
                                                    java.lang.reflect STILL NOT blocked
```

Narrowing the baseline alone is not sufficient. `java.lang` is unavoidable (every class extends `Object`), and under subtree matching it still admits `java.lang.reflect`, `java.lang.invoke`, `java.lang.instrument`, `java.lang.management`, `java.lang.foreign` and `java.lang.module`, precisely the capabilities governed by the separate reflection, agent-attach, environment, native-code and module-system rules. Fixing only the root moves the hole down one level.

Root cause layer: the architecture-enforcement layer (the `PACKAGE_IMPORT` rule and the allow-list that feeds it). It matters as a **false negative**: forbidden JDK packages were silently admitted, so student code could import capabilities a policy meant to deny.

- An instructor can now forbid an individual `java.*` package (for example `java.lang.reflect`) that was previously impossible to express, because the essential JDK entries are pinned to exact matches and the baseline is `java.lang` rather than all of `java`.
- The behaviour is fail-closed: an exercise must now grant the JDK packages it legitimately uses, rather than every JDK package being open by default.

- `PackagePermission` gains an `exactMatchOnly` component; the single-argument constructor keeps the subpackage-inclusive default, and Jackson binds policy YAML through the canonical constructor supplying the primitive default, so existing policies deserialise with unchanged meaning (pinned by a regression test against Jackson 2.22.0).
- Defence in depth: `JavaCreator` pins the JDK-namespace essentials (`java`, `javax`, `jdk`, `sun`, `com.sun`) to exact matches, so even if the YAML were widened back to a bare `java`, it would match only the literal package `java`, which declares no classes. Non-JDK essentials such as `org.aspectj` legitimately span subpackages and keep subtree matching.
- Both package-import outcome cache keys (`JavaWalaTestCase` and `JavaArchunitTestCase`) and both codegen sites now carry the flag, so a cached verdict or a generated test cannot silently apply the looser semantics.
- The rule previously had no behavioural coverage beyond its description string, which is why the bare `java` root survived. This adds 12 tests pinning the corrected behaviour.

## 4. Testing manual

**Prerequisites**

1. JDK 21 and Maven, as for any Ares 2 build. No exercise repository or echo server is needed; the added tests drive the real `PACKAGE_IMPORT` rule with local fixtures.
2. Check out this branch.

**Steps**

1. `mvn -o test -Dtest=PackageImportEnforcementTest -f pom.xml`

**Expected result**

- The 12 new tests pass, pinning: `java.io`/`java.net` blocked; `java.lang.reflect` blocked despite the `java.lang` baseline; collections and lambdas not allowed by default; plain `java.lang`-only code still passes; a widened essential list cannot re-open the JDK; policy subtree grants preserved; exact grants do not leak; boundary-awareness (`com.foo` does not cover `com.foobar`); ArchUnit/WALA verdict parity; legacy YAML back-compatibility.

- A widened essential list must not re-open the JDK: even a bare `java` essential entry matches only the literal package `java` (which declares no classes), so JDK packages stay forbidden unless a policy grants them.
- Boundary-awareness must hold: an exact grant of `com.foo` must not cover `com.foobar`.
- A policy subtree grant of a JDK package must still work (allow-list matching is a union, so a subtree entry wins over an exact entry): granting a JDK package as a subtree is an explicit, visible policy choice.

**Known failing tests (deliberately not fixed here).** Because `java` previously allowed the whole JDK, no policy in this repository grants any JDK package, so exercises must now grant what they use, which is the intended fail-closed behaviour but breaks the integration suite. Isolated against a clean-HEAD worktree:

| | Failures | Errors |
|---|---|---|
| HEAD | 149 | 472 |
| HEAD + this change | 150 | 1367 |

81 test classes, concentrated in `aop.allowed.*`, assert that permitted operations succeed but now trip `PACKAGE_IMPORT` first. The required grants are a bounded set across 42 policy files (`java.io` 936, `java.util.concurrent` 272, `java.util` 144, `java.nio.file` 96, `java.nio.channels` 64, `java.nio.charset` 24, `java.net` 17, `java.nio` 16, `java.awt` 8). Left as a follow-up so the enforcement fix can be reviewed on its own; this PR should stay draft until those policies are migrated.

**Modes exercised**

- [ ] ArchUnit + AspectJ
- [ ] WALA + AspectJ
- [ ] WALA + instrumentation

Instruction, Branch, Line and Method coverage for every production class changed in this PR. This branch's build is intentionally red (the integration suite fails until dependent policies are migrated, see "Known failing tests" in §4), so the "Coverage Report" job produced no aggregated JaCoCo report for this run; every counter is therefore `n/a` and will populate automatically once the branch is green. The 12 new `PackageImportEnforcementTest` tests and 106 unit tests across the touched areas pass, and static analysis (Checkstyle, PMD, SpotBugs) is clean.

| `PackagePermission` | n/a | n/a | n/a | n/a |
| `JavaArchunitTestCase` | n/a | n/a | n/a | n/a |
| `JavaArchunitTestCaseCollection` | n/a | n/a | n/a | n/a |
| `JavaWalaTestCase` | n/a | n/a | n/a | n/a |
| `JavaCreator` | n/a | n/a | n/a | n/a |

_All counters are `n/a` because no aggregated JaCoCo report was produced for this intentionally-red run. `EssentialPackages.yaml` is a resource, not a Java class._

**Yes, this is a breaking change for downstream exercise policies and warrants a release note and a major-version bump.** Because the JDK is no longer open by default, an exercise policy must now grant the `java.*` packages its permitted operations use (the bounded set enumerated in §4). Existing policy *files* deserialise with unchanged meaning: `PackagePermission`'s single-argument constructor keeps the subpackage-inclusive default and Jackson binds through the canonical constructor supplying the primitive default (pinned by a regression test). What changes is the effective allow-list, not the file format.

Two follow-ups identified while making this change, deliberately not fixed here:

- **Violation messages are weak.** 256 violations reported the package as literally `"unknown package"`. `parseErrorMessage` extracts the package by splitting on `.` until the first uppercase segment, falls back to that string when parsing fails, and only ever reports the first offending package. Now that package-import violations are expected, this is what exercise authors will rely on.
- **`allowedListedPackages` is dead config.** It is written by `JavaExecuter` but read nowhere in `src/`, so the essential list has no runtime AOP effect.

## Checklist

## Review progress

- [ ] Code review
- [ ] Manual test


src/main/resources/de/tum/cit/ase/ares/api/configuration/essentialFiles/java/EssentialPackages.yaml
```diff
@@ -1,5 +1,11 @@
+# SECURITY: keep this to the JDK packages that compiled code cannot avoid. Every
+# class extends java.lang.Object, so java.lang is unavoidable; everything else
+# (java.util, java.io, java.net, java.time, ...) is a deliberate choice and must be
+# granted explicitly via regardingPackageImports in the security policy. Entries in
+# the JDK namespace are granted as an exact match, so java.lang does NOT imply
+# java.lang.reflect, java.lang.invoke, java.lang.instrument or java.lang.management.
 essentialJavaPackages:
-  - java
+  - java.lang
```
Claudia-Anthropica, line 8: @MarkusPaulsen [high] Narrowing the baseline without migrating the bundled policies makes permitted student operations fail at `PACKAGE_IMPORT`. Both shipped examples import `java.io`, `java.nio`, and `java.util` while their `regardingPackageImports` lists remain empty, and the snapshot records failures for both examples, core integration, and all four enforcement modes; the unit test also still asserts that this list contains `java`. Migrate the bundled example and integration policies with the minimum packages required by each permitted scenario, update the stale essential-data test, and restore the affected jobs.

In `src/main/resources/de/tum/cit/ase/ares/api/configuration/essentialFiles/java/EssentialPackages.yaml`, narrowing the JDK baseline without migrating bundled policies causes false-positive package-import violations and breaks the examples and enforcement suites. Add the minimum required JDK package grants to every affected bundled policy, update the stale essential-package assertion, and verify all affected jobs pass.

</details>

src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/PackagePermission.java
```diff
@@ -16,10 +16,12 @@
...
  * @param importTheFollowingPackage the package that is permitted to be
- *                                  imported, or {@code *} for every package;
- *                                  must not be null.
+ *                                  imported; must not be null.
+ * @param exactMatchOnly            whether the permission covers only this
+ *                                  exact package ({@code true}) or the package
+ *                                  and its subpackages ({@code false}).
  */
-public record PackagePermission(@Nonnull String importTheFollowingPackage) {
+public record PackagePermission(@Nonnull String importTheFollowingPackage, boolean exactMatchOnly) {
```
Claudia-Anthropica, line 24: @MarkusPaulsen [medium] `exactMatchOnly` cannot be supplied through an actual security-policy YAML. `SecurityPolicySchemaValidator.PACKAGE_FIELDS` still permits only `importTheFollowingPackage`, so the production reader rejects this new key as unknown before Jackson reaches the canonical constructor; the added raw-`ObjectMapper` test bypasses that validator. Allow and type-check this field as an optional boolean in the strict schema and cover it through `SecurityPolicyYAMLReader`.

In `src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/PackagePermission.java`, the new `exactMatchOnly` component is rejected by the production policy schema validator, so YAML authors cannot request exact matching. Extend `SecurityPolicySchemaValidator` to allow and type-check `exactMatchOnly` as an optional boolean, and add a reader-level regression test.

</details>

MarkusPaulsen:
count towards it.

This description was written before those limits and exceeds 3 of them, so I have

<details><summary>1. Problem, as it read before (1544 characters, limit 1000)</summary>

Observed under both AOP modes and both architecture modes (WALA does not use its call graph for this category; it delegates to the same ArchUnit rule). `PACKAGE_IMPORT` could not forbid any `java.*` package.

The allow-list is assembled in `JavaCreator.prepareAllowedPackages` from `EssentialPackages.yaml`, which declared the bare root `java`. Matching was subpackage-inclusive (`startsWith`), so that one entry whitelisted the entire JDK. Reproduced against the real rule, with a class importing `java.io.File`, `java.net.Socket` and `java.lang.reflect.Method`:

```
allowed = [java, student]      (shipped config)  -> PASSED, nothing blocked
allowed = [java.lang, student]                   -> java.io + java.net blocked
                                                    java.lang.reflect STILL NOT blocked
```

Narrowing the baseline alone is not sufficient. `java.lang` is unavoidable (every class extends `Object`), and under subtree matching it still admits `java.lang.reflect`, `java.lang.invoke`, `java.lang.instrument`, `java.lang.management`, `java.lang.foreign` and `java.lang.module`, precisely the capabilities governed by the separate reflection, agent-attach, environment, native-code and module-system rules. Fixing only the root moves the hole down one level.

Root cause layer: the architecture-enforcement layer (the `PACKAGE_IMPORT` rule and the allow-list that feeds it). It matters as a **false negative**: forbidden JDK packages were silently admitted, so student code could import capabilities a policy meant to deny.

</details>

<details><summary>3. Improvement from the maintainer's perspective, as it read before (1088 characters, limit 1000)</summary>

- `PackagePermission` gains an `exactMatchOnly` component; the single-argument constructor keeps the subpackage-inclusive default, and Jackson binds policy YAML through the canonical constructor supplying the primitive default, so existing policies deserialise with unchanged meaning (pinned by a regression test against Jackson 2.22.0).
- Defence in depth: `JavaCreator` pins the JDK-namespace essentials (`java`, `javax`, `jdk`, `sun`, `com.sun`) to exact matches, so even if the YAML were widened back to a bare `java`, it would match only the literal package `java`, which declares no classes. Non-JDK essentials such as `org.aspectj` legitimately span subpackages and keep subtree matching.
- Both package-import outcome cache keys (`JavaWalaTestCase` and `JavaArchunitTestCase`) and both codegen sites now carry the flag, so a cached verdict or a generated test cannot silently apply the looser semantics.
- The rule previously had no behavioural coverage beyond its description string, which is why the bare `java` root survived. This adds 12 tests pinning the corrected behaviour.

</details>

<details><summary>Breaking changes and migration, as it read before (1244 characters, limit 1000)</summary>

**Yes, this is a breaking change for downstream exercise policies and warrants a release note and a major-version bump.** Because the JDK is no longer open by default, an exercise policy must now grant the `java.*` packages its permitted operations use (the bounded set enumerated in §4). Existing policy *files* deserialise with unchanged meaning: `PackagePermission`'s single-argument constructor keeps the subpackage-inclusive default and Jackson binds through the canonical constructor supplying the primitive default (pinned by a regression test). What changes is the effective allow-list, not the file format.

Two follow-ups identified while making this change, deliberately not fixed here:

- **Violation messages are weak.** 256 violations reported the package as literally `"unknown package"`. `parseErrorMessage` extracts the package by splitting on `.` until the first uppercase segment, falls back to that string when parsing fails, and only ever reports the first offending package. Now that package-import violations are expected, this is what exercise authors will rely on.
- **`allowedListedPackages` is dead config.** It is written by `JavaExecuter` but read nowhere in `src/`, so the essential list has no runtime AOP effect.

</details>

Claudia-Anthropica (changes requested):
@MarkusPaulsen Both existing blockers remain on this head. `SecurityPolicySchemaValidator.PACKAGE_FIELDS` still rejects `exactMatchOnly`, and narrowing the essential baseline to exact `java.lang` without migrating the bundled policies causes false-positive `PACKAGE_IMPORT` violations. The snapshot accordingly shows Unit Tests, Core Integration Tests, all four enforcement-mode jobs, and both example jobs failing.

PR #139, opened 2026-07-21
Configurable termination grace, structured network receiver check and a serialisation regression test

MarkusPaulsen:
## Summary

Three independent changes: make the post-interruption termination grace period configurable via `@StrictTimeout`, resolve the network receiver through the structured target path (with a fail-closed fix so an enforcement denial cannot be swallowed), and add a regression test pinning the British serialisation rule label.

## Linked issues

None.

## 1. Problem

Three separate items on one branch, each with its own root cause.

**Termination grace period was a fixed constant.** A timed-out test is interrupted once and then given a bounded period to leave its body before Ares treats the worker as contaminated and halts the fork. That period was a hard-coded 50 ms (`DEFAULT_TERMINATION_GRACE_PERIOD`), too short for interruption-aware cleanup in some exercises. Root cause layer: the timeout enforcement plumbing in `internal`/`api`. It matters as a false positive: a correct submission whose interruption-aware cleanup needs slightly longer than 50 ms is wrongly failed.

**The network receiver was checked only through the string-based criteria scan.** In the instrumentation network advice (`JavaInstrumentationAdviceNetworkSystemToolbox`), the receiver instance went through `checkIfVariableCriteriaIsViolated`, which reports a less precise display string than the structured parameter path. Separately, `variableToTarget` on a `URLConnection` called `getURL()` unconditionally, which can throw while the connection's constructor is still running (some JDK implementations delegate `getURL()` to an object not yet assigned). Root cause layer: the instrumentation (AOP) enforcement layer. The imprecise display string is a diagnostics weakness; the `getURL()` throw is a robustness gap.

**A subtle fail-open risk introduced while fixing the above.** The first cut of the `URLConnection` fix caught `RuntimeException` and returned `null` to tolerate the construction-time `getURL()` failure. A `SecurityException` is a `RuntimeException`, so an Ares enforcement denial raised while resolving the receiver would have been swallowed and turned into a `null` target, skipping the receiver check and failing open. This is the direction that matters for a security tool: it would let a forbidden receiver through.

**The serialisation rule label had no regression test.** The mapping that normalises the ArchUnit rule label `Serialises objects` to the action `serialise objects` was fixed upstream in `943a0cc0`; nothing pinned it, so a future casing leak into the reported sentence would go unnoticed.

- An instructor can now set `@StrictTimeout(terminationGrace = ..., terminationGraceUnit = ...)` so interruption-aware cleanup gets the time it needs, instead of being bounded to a fixed 50 ms. The defaults preserve the previous 50 ms, so nothing changes unless the attributes are set.
- A denied network interaction through the receiver is now reported with the more precise structured display string (for example `127.0.0.1:12345`) rather than the coarser criteria-scan string.
- A forbidden receiver can no longer slip through because an enforcement `SecurityException` was caught and turned into a `null` target during receiver resolution.

- The termination grace period is now read from the annotation in one place (`TimeoutUtils.findTerminationGracePeriod`), mirroring the existing `findTimeout`.
- The receiver check has a single resolved-target path: the dead criteria-scan fallback (which routed the same instance back through `variableToTarget`, having just returned `null`, so it could never throw) was removed, taking a Checkstyle formatting violation with it.
- New unit tests pin the previously uncovered behaviours: structured resolution of a connected `DatagramSocket` and `DatagramChannel`, the `null` return when a `URLConnection` cannot expose its URL during construction, the instructor-configured grace period, bounded-cleanup completion, and the normalised serialisation label (including that the capitalised original label does not leak).

**Known gap, deliberately not addressed here:** `JqwikStrictTimeoutExtension` passes its own hard-coded one-second period to the three-argument overload, so the new `@StrictTimeout` attributes do not affect jqwik properties yet.

**Dropped from an earlier revision:** a commit that also disabled `StrictTimeout` for the JUnit lifecycle hooks was removed. It reproducibly broke `StrictTimeoutTest.test_time_testClassFailLoop` and would have left student code invoked from an instructor `@BeforeEach` unbounded.

## 4. Testing manual

**Prerequisites**

1. JDK 21 and Maven, as for any Ares 2 build. No exercise repository or policy file is needed; the added tests drive the advice and the timeout utilities directly. A loopback address is used by the datagram tests, no external echo server is required.
2. Check out this branch.

**Steps**

1. `mvn -o test -Dtest=TimeoutUtilsTest,JavaInstrumentationAdviceNetworkSystemToolboxTest,JavaArchitectureTestCaseTest -f pom.xml`
2. Full run for the mode combinations: `mvn test -f pom.xml`

**Expected result**

- Step 1: the targeted tests pass, including `resolvesInstructorConfiguredTerminationGracePeriod` (resolves 10 s from `@StrictTimeout(value = 20, unit = MILLISECONDS, terminationGrace = 10, terminationGraceUnit = SECONDS)`), `interruptionAwareExecutionHasTimeForBoundedFrameworkCleanup`, `toTarget_extractsRemoteHostAndPortFromConnectedDatagramSocket` / `...DatagramChannel` (both display `127.0.0.1:12345`), `toTarget_returnsNullWhenUrlConnectionCannotExposeUrlDuringConstruction`, and `testParseErrorMessage_withBritishSerializationRule_normalisesAction`.
- Step 2: `BUILD SUCCESS`. The full branch run was 2619 tests, 0 failures, 0 errors, 230 skipped.

- An untrusted runtime type is still rejected: `requireTrustedRuntimeType` stays outside the guarded block, so it is not swallowed by the `getURL()` tolerance.
- An enforcement `SecurityException` raised while resolving a `URLConnection` receiver must still propagate (fail closed), not be caught and turned into a `null` target: the `catch (SecurityException denied) { throw denied; }` before the broad `catch` guarantees this.
- The attribute scan stays unconditional: a target resolved from parameters or receiver must not suppress the attribute check, because attributes are a disjoint value set.
- The normalised serialisation message must contain `serialise objects` and must not leak the capitalised `Serialises objects` label.

**Modes exercised**

- [x] ArchUnit + AspectJ
- [x] ArchUnit + instrumentation
- [x] WALA + AspectJ
- [x] WALA + instrumentation

Instruction, Branch, Line and Method coverage for every production class changed in this PR, read from the aggregated JaCoCo report produced by the "Coverage Report" job of this PR's Maven workflow run. Nested classes (for example a record's `Builder`) are listed as their own rows. `n/a` means the class has no executable lines of that counter (for example an annotation or interface) or was not present in the report.

| `TimeoutUtils` | 86.9% (245/282) | 83.3% (20/24) | 85.5% (59/69) | 94.1% (16/17) |
| `TimeoutUtils.WhitelistedThreadFactory` | 88.5% (23/26) | 50.0% (1/2) | 83.3% (5/6) | 100.0% (2/2) |
| `JavaInstrumentationAdviceNetworkSystemToolbox` | 70.5% (991/1405) | 65.4% (189/289) | 73.2% (227/310) | 91.7% (22/24) |
| `StrictTimeout` | n/a | n/a | n/a | n/a |
| **Total (changed classes)** | **73.5% (1259/1713)** | **66.7% (210/315)** | **75.6% (291/385)** | **93.0% (40/43)** |

_`StrictTimeout` is an annotation type with no executable lines; its two new attributes are exercised via `TimeoutUtilsTest.resolvesInstructorConfiguredTerminationGracePeriod`._

None. `@StrictTimeout` gains `terminationGrace` and `terminationGraceUnit`, both with defaults (`50` and `TimeUnit.MILLISECONDS`) that preserve the previous behaviour, so existing annotations are unaffected. This is an additive public-API change under `de.tum.cit.ase.ares.api`, not a break. No policy file format, generated security test code or minimum JDK/Maven/Gradle version changes.

## Checklist

- [x] Documentation (`docs/`, `README.adoc`, Javadoc) was updated where the change is user-facing. The new `@StrictTimeout` attributes carry Javadoc; no `docs/` page describes the fixed-50 ms grace period.

## Review progress

- [ ] Code review
- [ ] Manual test

### DIFF
--- a/src/main/java/de/tum/cit/ase/ares/api/StrictTimeout.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/StrictTimeout.java
@@ -47,4 +47,15 @@ public @interface StrictTimeout {
 	 * @see TimeUnit
 	 */
 	TimeUnit unit() default TimeUnit.SECONDS;
+
+	/**
+	 * The bounded period allowed for an interrupted test to terminate before Ares
+	 * treats the test worker as contaminated.
+	 */
+	long terminationGrace() default 50;
+
+	/**
+	 * The unit of the termination grace period, <b>defaults to milliseconds</b>.
+	 */
+	TimeUnit terminationGraceUnit() default TimeUnit.MILLISECONDS;
 }
--- a/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
@@ -293,7 +293,14 @@ public final class JavaInstrumentationAdviceNetworkSystemToolbox extends JavaIns
 		}
 		if (value instanceof URLConnection urlConnection) {
 			requireTrustedRuntimeType(value);
-			return variableToTarget(urlConnection.getURL());
+			try {
+				return variableToTarget(urlConnection.getURL());
+			} catch (RuntimeException ignored) {
+				// Some JDK URLConnection implementations delegate getURL() to an object
+				// that is not assigned until construction has completed. Advice may observe
+				// such an instance while its constructor is still running.
+				return null;
+			}
 		}
 		if (value instanceof DatagramPacket datagramPacket) {
 			requireTrustedRuntimeType(value);
@@ -843,15 +850,28 @@ public final class JavaInstrumentationAdviceNetworkSystemToolbox extends JavaIns
 		// </editor-fold>
 		// <editor-fold desc="Check receiver instance">
 		@Nullable
-		String networkIllegallyInteractedThroughReceiver = instance == null ? null
-				: checkIfVariableCriteriaIsViolated(new Object[] { instance }, allowedHosts, allowedPorts,
-						IgnoreValues.NONE);
-		if (networkIllegallyInteractedThroughReceiver != null) {
-			throw new SecurityException(localize("security.advice.illegal.network.execution",
-					networkSystemMethodToCheck, action, networkIllegallyInteractedThroughReceiver,
-					fullMethodSignature
-							+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
-							+ buildDenialReason(noAllowRuleConfigured)));
+		NetworkTarget targetFromReceiver = variableToTarget(instance);
+		if (targetFromReceiver != null) {
+			if (checkIfNetworkIsForbidden(targetFromReceiver, allowedHosts, allowedPorts)) {
+				throw new SecurityException(localize("security.advice.illegal.network.execution",
+						networkSystemMethodToCheck, action, targetFromReceiver.toDisplayString(),
+						fullMethodSignature
+								+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
+								+ buildDenialReason(noAllowRuleConfigured)));
+			}
+		} else if (instance != null) {
+			// Fall back to the criteria scan for receivers that cannot be resolved into a
+			// structured target, so no receiver type loses its check.
+			@Nullable
+			String networkIllegallyInteractedThroughReceiver = checkIfVariableCriteriaIsViolated(
+					new Object[] { instance }, allowedHosts, allowedPorts, IgnoreValues.NONE);
+			if (networkIllegallyInteractedThroughReceiver != null) {
+				throw new SecurityException(localize("security.advice.illegal.network.execution",
+						networkSystemMethodToCheck, action, networkIllegallyInteractedThroughReceiver,
+						fullMethodSignature
+								+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
+								+ buildDenialReason(noAllowRuleConfigured)));
+			}
 		}
 		// </editor-fold>
 		// <editor-fold desc="Check attributes">
--- a/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
@@ -851,27 +851,16 @@ public final class JavaInstrumentationAdviceNetworkSystemToolbox extends JavaIns
 		// <editor-fold desc="Check receiver instance">
 		@Nullable
 		NetworkTarget targetFromReceiver = variableToTarget(instance);
-		if (targetFromReceiver != null) {
-			if (checkIfNetworkIsForbidden(targetFromReceiver, allowedHosts, allowedPorts)) {
-				throw new SecurityException(localize("security.advice.illegal.network.execution",
-						networkSystemMethodToCheck, action, targetFromReceiver.toDisplayString(),
-						fullMethodSignature
-								+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
-								+ buildDenialReason(noAllowRuleConfigured)));
-			}
-		} else if (instance != null) {
-			// Fall back to the criteria scan for receivers that cannot be resolved into a
-			// structured target, so no receiver type loses its check.
-			@Nullable
-			String networkIllegallyInteractedThroughReceiver = checkIfVariableCriteriaIsViolated(
-					new Object[] { instance }, allowedHosts, allowedPorts, IgnoreValues.NONE);
-			if (networkIllegallyInteractedThroughReceiver != null) {
-				throw new SecurityException(localize("security.advice.illegal.network.execution",
-						networkSystemMethodToCheck, action, networkIllegallyInteractedThroughReceiver,
-						fullMethodSignature
-								+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
-								+ buildDenialReason(noAllowRuleConfigured)));
-			}
+		// No separate criteria scan is needed for a receiver that does not resolve into
+		// a structured target: the scan would route this same instance back through
+		// variableToTarget, which just returned null, so it could never throw. Only the
+		// resolved-target check below can flag a forbidden receiver.
+		if (targetFromReceiver != null && checkIfNetworkIsForbidden(targetFromReceiver, allowedHosts, allowedPorts)) {
+			throw new SecurityException(localize("security.advice.illegal.network.execution",
+					networkSystemMethodToCheck, action, targetFromReceiver.toDisplayString(),
+					fullMethodSignature
+							+ (studentCalledMethod == null ? "" : " (called by " + studentCalledMethod + ")") + " | "
+							+ buildDenialReason(noAllowRuleConfigured)));
 		}
 		// </editor-fold>
 		// <editor-fold desc="Check attributes">
--- a/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
@@ -295,6 +295,11 @@ public final class JavaInstrumentationAdviceNetworkSystemToolbox extends JavaIns
 			requireTrustedRuntimeType(value);
 			try {
 				return variableToTarget(urlConnection.getURL());
+			} catch (SecurityException denied) {
+				// A SecurityException here is an Ares enforcement denial, not a resolution
+				// failure. Swallowing it would turn a denial into a null target and skip the
+				// receiver check, so it must propagate rather than fail open.
+				throw denied;
 			} catch (RuntimeException ignored) {
 				// Some JDK URLConnection implementations delegate getURL() to an object
 				// that is not assigned until construction has completed. Advice may observe
--- a/src/main/java/de/tum/cit/ase/ares/api/internal/TimeoutUtils.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/internal/TimeoutUtils.java
@@ -30,8 +30,14 @@ public final class TimeoutUtils {
 		return strictTimeout.map(st -> Duration.of(st.value(), st.unit().toChronoUnit()));
 	}
 
+	static Optional<Duration> findTerminationGracePeriod(TestContext context) {
+		var strictTimeout = TestContextUtils.findAnnotationIn(context, StrictTimeout.class);
+		return strictTimeout.map(st -> Duration.of(st.terminationGrace(), st.terminationGraceUnit().toChronoUnit()));
+	}
+
 	public static <T> T performTimeoutExecution(ThrowingSupplier<T> execution, TestContext context) throws Throwable {
-		return performTimeoutExecution(execution, context, DEFAULT_TERMINATION_GRACE_PERIOD);
+		Duration terminationGracePeriod = findTerminationGracePeriod(context).orElse(DEFAULT_TERMINATION_GRACE_PERIOD);
+		return performTimeoutExecution(execution, context, terminationGracePeriod);
 	}
 
 	public static <T> T performTimeoutExecution(ThrowingSupplier<T> execution, TestContext context,
--- a/src/test/java/de/tum/cit/ase/ares/api/architecture/java/JavaArchitectureTestCaseTest.java
+++ b/src/test/java/de/tum/cit/ase/ares/api/architecture/java/JavaArchitectureTestCaseTest.java
@@ -106,4 +106,17 @@ public class JavaArchitectureTestCaseTest {
 						|| thrown.getMessage().contains("Ares Sicherheitsfehler"),
 				"Exception message should contain Ares Security Error prefix");
 	}
+
+	@Test
+	void testParseErrorMessage_withBritishSerializationRule_normalisesAction() {
+		String message = "Architecture Violation [Priority: MEDIUM] - Rule 'Serialises objects' was violated (1 time):\n"
+				+ "Method <com.example.Test.serialize()> calls method <java.io.ObjectOutputStream.writeObject(java.lang.Object)>";
+		AssertionError error = new AssertionError(message);
+
+		SecurityException thrown = assertThrows(SecurityException.class,
+				() -> JavaArchitectureTestCase.parseErrorMessage(error));
+
+		assertTrue(thrown.getMessage().contains("serialise objects"),
+				() -> "Exception message should use the normalised serialisation action: " + thrown.getMessage());
+	}
 }
--- a/src/test/java/de/tum/cit/ase/ares/api/architecture/java/JavaArchitectureTestCaseTest.java
+++ b/src/test/java/de/tum/cit/ase/ares/api/architecture/java/JavaArchitectureTestCaseTest.java
@@ -118,5 +118,7 @@ public class JavaArchitectureTestCaseTest {
 
 		assertTrue(thrown.getMessage().contains("serialise objects"),
 				() -> "Exception message should use the normalised serialisation action: " + thrown.getMessage());
+		assertFalse(thrown.getMessage().contains("Serialises objects"),
+				() -> "Exception message must not leak the original capitalised rule label: " + thrown.getMessage());
 	}
 }
--- a/src/test/java/de/tum/cit/ase/ares/api/internal/TimeoutUtilsTest.java
+++ b/src/test/java/de/tum/cit/ase/ares/api/internal/TimeoutUtilsTest.java
@@ -58,6 +58,40 @@ class TimeoutUtilsTest {
 	}
 
 	@Test
+	void interruptionAwareExecutionHasTimeForBoundedFrameworkCleanup() throws Exception {
+		AtomicBoolean workerFinished = new AtomicBoolean();
+		AtomicInteger requestedExitCode = new AtomicInteger(-1);
+		TestContext context = contextFor("strictTimeoutTarget"); //$NON-NLS-1$
+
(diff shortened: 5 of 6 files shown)


src/test/java/de/tum/cit/ase/ares/api/architecture/java/JavaArchitectureTestCaseTest.java
```diff
@@ -106,4 +106,17 @@ void testParseErrorMessage_withPackageImportViolation_extractsForbiddenPackages(
...
+	void testParseErrorMessage_withBritishSerializationRule_normalisesAction() {
+		String message = "Architecture Violation [Priority: MEDIUM] - Rule 'Serialises objects' was violated (1 time):\n"
+				+ "Method <com.example.Test.serialize()> calls method <java.io.ObjectOutputStream.writeObject(java.lang.Object)>";
+		AssertionError error = new AssertionError(message);
+
+		SecurityException thrown = assertThrows(SecurityException.class,
+				() -> JavaArchitectureTestCase.parseErrorMessage(error));
+
+		assertTrue(thrown.getMessage().contains("serialise objects"),
+				() -> "Exception message should use the normalised serialisation action: " + thrown.getMessage());
```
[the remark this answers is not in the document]
MarkusPaulsen: Added the negative assertion: `assertFalse(thrown.getMessage().contains("Serialises objects"))`. The normalised message is built only from the mapped action ("serialise objects"), never the original rule label, so a future casing leak now fails the test.

src/main/java/de/tum/cit/ase/ares/api/internal/TimeoutUtils.java
```diff
@@ -30,8 +30,14 @@ public static Optional<Duration> findTimeout(TestContext context) {
...
+	static Optional<Duration> findTerminationGracePeriod(TestContext context) {
+		var strictTimeout = TestContextUtils.findAnnotationIn(context, StrictTimeout.class);
+		return strictTimeout.map(st -> Duration.of(st.terminationGrace(), st.terminationGraceUnit().toChronoUnit()));
+	}
+
 	public static <T> T performTimeoutExecution(ThrowingSupplier<T> execution, TestContext context) throws Throwable {
-		return performTimeoutExecution(execution, context, DEFAULT_TERMINATION_GRACE_PERIOD);
+		Duration terminationGracePeriod = findTerminationGracePeriod(context).orElse(DEFAULT_TERMINATION_GRACE_PERIOD);
+		return performTimeoutExecution(execution, context, terminationGracePeriod);
 	}
```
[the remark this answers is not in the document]
MarkusPaulsen: Leaving the two lookups as they are. The overloads are deliberately layered: the two-argument entry point derives the termination grace, while the package-private four-argument overload derives the timeout and is itself a direct entry point the tests call with an explicit grace and process terminator. Collapsing the two StrictTimeout resolutions would mean threading the resolved annotation (or both durations) through those signatures, coupling the layers for a single reflective `findAnnotationIn` that runs once per test method on a cold path. The clarity of the overload chain is worth more than avoiding that one lookup.

src/test/java/de/tum/cit/ase/ares/api/internal/TimeoutUtilsTest.java
```diff
@@ -57,6 +57,40 @@ void interruptionAwareExecutionFinishesBeforeControlReturns() throws Exception {
...
+				while (!Thread.currentThread().isInterrupted()) {
+					Thread.onSpinWait();
+				}
+			} finally {
+				long cleanupDeadline = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(200);
+				while (System.nanoTime() < cleanupDeadline) {
+					Thread.onSpinWait();
+				}
+				workerFinished.set(true);
+			}
```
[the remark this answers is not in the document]
MarkusPaulsen: Keeping the busy-spin. It deliberately avoids park/sleep so the test does not have to clear and restore the interrupt flag the timed-out worker already carries: the fixture models interruption-aware cleanup that runs to completion within the grace, and a time-based park would immediately throw InterruptedException because the interrupt is set. The 200 ms is deterministic and an order of magnitude inside the one-second grace, so it neither flakes nor meaningfully loads the suite; the interrupt-flag juggling a park would need is more fragile than the tight loop.

src/main/java/de/tum/cit/ase/ares/api/StrictTimeout.java
```diff
@@ -47,4 +47,15 @@
 	 * @see TimeUnit
 	 */
 	TimeUnit unit() default TimeUnit.SECONDS;
+
+	/**
+	 * The bounded period allowed for an interrupted test to terminate before Ares
+	 * treats the test worker as contaminated.
+	 */
+	long terminationGrace() default 50;
```
Claudia-Anthropica: @MarkusPaulsen [medium] `terminationGrace` is not honoured for jqwik properties: `JqwikStrictTimeoutExtension` still calls the three-argument overload with its hard-coded one-second grace. Consequently, an instructor who configures a longer grace here can still have a correct interruption-aware jqwik execution terminated after one second.

In `src/main/java/de/tum/cit/ase/ares/api/StrictTimeout.java`, the new termination-grace override is ignored by `JqwikStrictTimeoutExtension`, which always uses its hard-coded one-second grace. Make the jqwik path consume the configured annotation value while retaining its legacy one-second fallback when no override is configured, and add a jqwik regression test.

</details>

MarkusPaulsen: Applied in `fdc6ebbe`, though not as suggested, and the difference is the whole
point.

The finding is correct: `JqwikStrictTimeoutExtension` passed its hard-coded second
as the period to use, so a configured `terminationGrace` never reached a property.

The suggested fix, consume the annotation while retaining the legacy one-second
fallback, would have made things worse. `terminationGrace` defaulted to `50`, so
`findTerminationGracePeriod` could not tell "not configured" from "configured to
50 ms". The fallback would then have fired only when no `@StrictTimeout` was in
scope at all, and in that case `findTimeout` is empty and no timeout runs, so the
fallback was unreachable in practice. Every existing jqwik property carrying
`@StrictTimeout` would have dropped from a second to 50 ms: a false-positive
regression, in the direction this pull request set out to remove.

What was actually needed was for the annotation to have an unset state.
`terminationGrace` now defaults to `-1`, `findTerminationGracePeriod` returns
empty for a negative value, and the third parameter of `performTimeoutExecution`
became the caller's *default* rather than the period to use, resolved in one place
for both extensions. Jupiter keeps 50 ms, jqwik keeps a second, and a configured
value beats either. Zero stays a configured value meaning "terminate at once".

Two further defects fell out of making the period instructor-controlled, neither
reachable while it was a constant, both in `TimeoutUtils`:

- A sub-millisecond period was rounded down by `Duration.toMillis()` to the zero
  that means terminate at once.
- A period large enough to overflow `toMillis()` threw an `ArithmeticException`
  out of `terminateTimedOutExecution`, after the worker had been cancelled but
  **before** the fatal terminator ran. A fork whose untrusted code ignores
  interruption would have survived. That is a fail-open, and it is why this is
  not a wiring change.

Every period, from the annotation or from the caller, is now bounds-checked and
converted to nanoseconds before the worker is created, so nothing is converted
after cancellation.

On the requested jqwik regression test: a live property that overruns its grace
is deliberately not the test, because the behaviour under test halts the JVM, so
a regression would take the surefire fork down rather than report a failure. The
contract is pinned instead by
`JqwikStrictTimeoutExtensionTest.contributesTheOneSecondDefaultTerminationGracePeriod`
and `TimeoutUtilsTest.jupiterKeepsItsFiftyMillisecondDefault`, which pin the two
defaults separately, together with `configuredTerminationGraceOverridesTheCallersDefault`
on the one method both extensions go through. Six further tests cover the unset,
zero, sub-millisecond, over-bound and unusable-caller-default cases.

60 tests pass: `TimeoutUtilsTest` 13, `JqwikStrictTimeoutExtensionTest` 1,
`StrictTimeoutTest` 13, `JavaInstrumentationAdviceNetworkSystemToolboxTest` 23,
`JavaArchitectureTestCaseTest` 10. The pull request description has been updated,
including the breaking-changes section, which now records the two internal
signature changes and the new rejection.

Claudia-Anthropica (dismissed):
@MarkusPaulsen The jqwik path now honours configured termination grace while preserving its one-second fallback, and grace periods are validated and converted before worker execution. The network denial remains fail-closed, all prior findings are resolved, and the build, static analysis, and all four integration-mode checks succeeded; only the non-required unit-test check was still running in the snapshot.

MarkusPaulsen:
@coderabbitai review

Nothing here is outstanding as far as I can tell, so this is a request for a fresh
pass rather than a response to open findings.

The two inline findings from the 2026-07-21 review are resolved. The one **outside
diff range comment** from 2026-07-27, not swallowing `SecurityException` from
`getURL()`, is implemented in
[`JavaInstrumentationAdviceNetworkSystemToolbox.java:298`](https://github.com/ls1intum/Ares2/blob/fix/timeout-grace-network-advice-and-arch-message/src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java#L298):
the `catch (SecurityException denied) { throw denied; }` sits before the broad
`catch`, so an enforcement denial propagates instead of becoming a `null` target.

Since then `fdc6ebbe` closed the jqwik termination-grace gap. That commit was never
reviewed here: the `CodeRabbit` commit status on the head reads `Review rate
limited`, so no incremental pass ran on it. Worth a look, because it changes an
enforcement-adjacent path rather than only wiring one up. It also fixes a fail-open
found while doing so: a large instructor-configured grace made `Duration.toMillis()`
throw out of `terminateTimedOutExecution` after the worker was cancelled but before
the fatal terminator ran, which would have left a contaminated fork alive. Every
period is now bounds-checked and converted to nanoseconds before the worker is
created.

All 15 checks are green and every review thread is resolved.

MarkusPaulsen:
count towards it.

This description was written before those limits and exceeds 3 of them, so I have

<details><summary>1. Problem, as it read before (2142 characters, limit 1000)</summary>

Three separate items on one branch, each with its own root cause.

**Termination grace period was a fixed constant.** A timed-out test is interrupted once and then given a bounded period to leave its body before Ares treats the worker as contaminated and halts the fork. That period was a hard-coded 50 ms (`DEFAULT_TERMINATION_GRACE_PERIOD`), too short for interruption-aware cleanup in some exercises. Root cause layer: the timeout enforcement plumbing in `internal`/`api`. It matters as a false positive: a correct submission whose interruption-aware cleanup needs slightly longer than 50 ms is wrongly failed.

**The network receiver was checked only through the string-based criteria scan.** In the instrumentation network advice (`JavaInstrumentationAdviceNetworkSystemToolbox`), the receiver instance went through `checkIfVariableCriteriaIsViolated`, which reports a less precise display string than the structured parameter path. Separately, `variableToTarget` on a `URLConnection` called `getURL()` unconditionally, which can throw while the connection's constructor is still running (some JDK implementations delegate `getURL()` to an object not yet assigned). Root cause layer: the instrumentation (AOP) enforcement layer. The imprecise display string is a diagnostics weakness; the `getURL()` throw is a robustness gap.

**A subtle fail-open risk introduced while fixing the above.** The first cut of the `URLConnection` fix caught `RuntimeException` and returned `null` to tolerate the construction-time `getURL()` failure. A `SecurityException` is a `RuntimeException`, so an Ares enforcement denial raised while resolving the receiver would have been swallowed and turned into a `null` target, skipping the receiver check and failing open. This is the direction that matters for a security tool: it would let a forbidden receiver through.

**The serialisation rule label had no regression test.** The mapping that normalises the ArchUnit rule label `Serialises objects` to the action `serialise objects` was fixed upstream in `943a0cc0`; nothing pinned it, so a future casing leak into the reported sentence would go unnoticed.

</details>

<details><summary>3. Improvement from the maintainer's perspective, as it read before (2986 characters, limit 1000)</summary>

- The termination grace period is now read from the annotation in one place (`TimeoutUtils.findTerminationGracePeriod`), mirroring the existing `findTimeout`.
- The receiver check has a single resolved-target path: the dead criteria-scan fallback (which routed the same instance back through `variableToTarget`, having just returned `null`, so it could never throw) was removed, taking a Checkstyle formatting violation with it.
- New unit tests pin the previously uncovered behaviours: structured resolution of a connected `DatagramSocket` and `DatagramChannel`, the `null` return when a `URLConnection` cannot expose its URL during construction, the instructor-configured grace period, bounded-cleanup completion, and the normalised serialisation label (including that the capitalised original label does not leak).

**Closed after review:** the jqwik path was left out of the first revision, on the grounds that `JqwikStrictTimeoutExtension` passes its own period to the three-argument overload. It is now included, and closing it needed one design change rather than a wiring change.

`terminationGrace` defaulted to `50`, so "not configured" and "configured to 50 ms" were the same value. The two extensions do not share one period, Jupiter allowing 50 ms and jqwik a second because a try unwinds through the property lifecycle as well as the test body, and no single attribute default can express both. Consuming the attribute as it stood would therefore have cut every existing jqwik property from a second to 50 ms, which is the false-positive direction this pull request set out to remove. The attribute now defaults to a negative value meaning unset, and the third parameter of `performTimeoutExecution` is the caller's *default* rather than the period to use, resolved in one place for both extensions.

Making the period instructor-controlled also made two conversions reachable that a hard-coded constant never exercised, both in `TimeoutUtils`:

- A sub-millisecond period was rounded down by `Duration.toMillis()` to the zero that means "terminate at once".
- A period large enough to overflow `toMillis()` threw an `ArithmeticException` out of `terminateTimedOutExecution` after the worker had been cancelled but **before** the fatal terminator ran, leaving a fork whose untrusted code ignores interruption alive. That is a fail-open, and it is the reason this is not merely a wiring change.

Every period, from the annotation or from the caller, is now bounds-checked and converted to nanoseconds before the worker is created, so no conversion happens after cancellation at all. The bound is one day, far beyond any period a test can want and well inside what the wait's arithmetic holds.

**Dropped from an earlier revision:** a commit that also disabled `StrictTimeout` for the JUnit lifecycle hooks was removed. It reproducibly broke `StrictTimeoutTest.test_time_testClassFailLoop` and would have left student code invoked from an instructor `@BeforeEach` unbounded.

</details>

<details><summary>Breaking changes and migration, as it read before (1315 characters, limit 1000)</summary>

None for an existing annotation. `@StrictTimeout` gains `terminationGrace` and `terminationGraceUnit`. `terminationGrace` defaults to `-1`, meaning "not configured", which leaves each extension its own previous period, 50 ms for JUnit and one second for jqwik. An annotation written before this change therefore behaves exactly as it did. This is an additive public-API change under `de.tum.cit.ase.ares.api`, not a break. No policy file format, generated security test code or minimum JDK/Maven/Gradle version changes.

Two internal signatures in `TimeoutUtils` changed shape, both `@API(status = INTERNAL)`: `terminateTimedOutExecution` and the package-private `executeWithTimeout` take a `long` nanosecond count instead of a `Duration`, so that no conversion can throw after the worker has been cancelled. The public `performTimeoutExecution` overloads keep their signatures; the third argument's meaning changed from "the period to use" to "the period to use when the annotation configures none", which affects the two extensions in this repository and no other caller.

One new rejection: a `terminationGrace` above one day, or one no `Duration` can hold, is refused with an `IllegalArgumentException` when the annotation is read, before the test runs. No such annotation exists, and one could not have worked.

</details>

src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
```diff
@@ -424,6 +421,8 @@ private static boolean carriesNoEndpoint(@Nonnull SocketAddress socketAddress,
 	private static URL urlOfConnection(@Nonnull URLConnection urlConnection) {
 		try {
 			return urlConnection.getURL();
+		} catch (SecurityException denied) {
```
Claudia-Anthropica: @MarkusPaulsen [medium] This rethrow is the fail-closed fix, but no test makes `getURL()` throw a `SecurityException` and verifies that it propagates; the added `URLConnection` test covers only the following null-return branch. Because losing this catch ordering converts an enforcement denial into an allowed unresolved target, add a regression test for this branch.

In `src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java`, the new `SecurityException` rethrow branch has no regression coverage. Add a unit test that invokes `urlOfConnection` with a `URLConnection` whose `getURL()` throws `SecurityException` and asserts that exception remains the cause rather than becoming `null`.

</details>

Claudia-Anthropica (changes requested):
@MarkusPaulsen The jqwik path now honours configured grace while retaining both legacy defaults, and the URL resolver explicitly preserves `SecurityException`; the earlier functional findings are fixed. One security-relevant regression test is still missing: the current `URLConnection` test exercises only the broad `RuntimeException`-to-null path and would not detect restoration of the original denial-swallowing behaviour.

src/main/java/de/tum/cit/ase/ares/api/aop/java/instrumentation/advice/JavaInstrumentationAdviceNetworkSystemToolbox.java
MarkusPaulsen: Fixed in b18a1e64. Both backends now rethrow a SecurityException from getURL() before the broad RuntimeException catch, and each has a regression test that invokes urlOfConnection with a connection whose getURL throws a denial and asserts it propagates (urlOfConnection_letsADenialThrough in JavaInstrumentationAdviceNetworkSystemToolboxTest and in JavaAspectJNetworkSystemAdviceDefinitionsTest). Both tests fail if the catch order is lost. The AspectJ twin had the same swallowing behaviour and is fixed with it.

Claudia-Anthropica (dismissed):
@MarkusPaulsen Both network backends now rethrow `SecurityException` from `URLConnection.getURL()`, and their regression tests close the prior fail-open coverage gap. The jqwik path honours configured grace while preserving its legacy default, with validation completed before worker execution. No unresolved findings remain; the build and PR-template checks succeeded, while the remaining captured jobs were non-required and still in progress.

Claudia-Anthropica (approved):
@MarkusPaulsen The exact head preserves both timeout defaults while allowing configured grace periods, validating and converting them before worker execution. Both URL backends keep enforcement denials fail-closed with regression coverage, and the structured receiver and serialisation changes are consistent with the surrounding code. All captured checks succeeded, and no unresolved findings remain.

PR #141, opened 2026-07-21
Report aggregated code coverage in CI

MarkusPaulsen:
## Summary

Activates the long-dormant `coverage` profile in CI and adds a `Coverage Report` job that merges the JaCoCo execution data of all seven test jobs into one aggregated report. Thresholds are advisory for now, because the project has never measured its own coverage.

## Linked issues

None.

## 1. Problem

The `coverage` profile has existed in `pom.xml` for a long time, complete with a BUNDLE-level `jacoco:check` rule of 55% lines, 35% branches and 65% methods. Nothing ever ran it: no workflow passes `-Pcoverage`, and no workflow reaches the `verify` phase, which is where the check is bound. The rule was therefore dead configuration, and the project had no coverage figures whatsoever, neither for a pull request nor for the repository as a whole.

The obvious fix, adding one coverage job, does not work here. CI deliberately splits the suite across seven jobs (`Unit Tests`, the four-way integration matrix, `Core Integration Tests`), each covering a disjoint slice. Any single slice measured on its own falls far below the BUNDLE minima and would fail a rule that is in fact working correctly. Only the union of the slices can be judged against it.

This is not a defect in the enforcement layer, so neither a false negative nor a false positive. It is a maintenance gap in the build integration.

No Improvement. Nothing an instructor or student observes at runtime changes: no policy behaviour, no generated test code, and no public API.

Every pull request now gets an aggregated coverage table in the job summary of the `Coverage Report` job, plus a `coverage-report` artefact with the full HTML and CSV report, so per-class line coverage can be quoted in section 5 of this template without running anything locally. The threshold rule stops being decorative, and the two places that state the minima are collapsed into shared properties, so the local full-suite gate and the CI aggregate cannot drift apart.

It also documents a real trap: JaCoCo must not instrument the integration test subjects, see below.

## 4. Testing manual

**Prerequisites**

1. A checkout of this branch, JDK 21, Maven. No exercise repository and no echo server are needed.

**Steps**

Not reproducible from an exercise; this is a CI and build configuration change. A reviewer verifies it as follows.

1. Open the `Coverage Report` job of the Maven workflow run on this pull request.
2. Read the aggregated table in the job summary.
3. Download the `coverage-report` artefact and open `index.html` from it.
4. Confirm the six `jacoco-exec-*` artefacts exist on the run, one per test job.
5. Locally, reproduce a single slice and the aggregation:
   ```
   mvn test -Punit-core-tests,coverage -f pom.xml
   mvn verify -Pcoverage-aggregate -DskipTests -Djacoco.haltOnFailure=false -f pom.xml
   ```

**Expected result**

The summary shows one row each for instruction, branch, line and method coverage. Where a threshold is not met, the job summary additionally lists the `Rule violated` lines, and the job still succeeds, because the check is advisory for the moment.

Ares must not become more permissive by being measured. Every forbidden-access test must keep failing the student code exactly as before, in particular the architecture tests, since attaching a second agent is precisely what perturbed them. Confirm that all four matrix jobs are green on this pull request, and that `Tests run` per job matches a run on `main`. Locally, the ArchUnit/AspectJ selection of 302 tests passes with `-Pcoverage` and without it, with identical results.

**Modes exercised**

- [x] ArchUnit + AspectJ
- [ ] WALA + AspectJ
- [ ] WALA + instrumentation

ArchUnit + AspectJ was verified locally, because that is the combination whose three architecture tests broke when the agent instrumented the test subjects, and the same 302-test selection now passes with and without coverage. The remaining three combinations are exercised by this pull request's own CI run, which is the same code path with a different method selection.

No Java code changed.

None.

## Checklist

- [ ] CI is green, or every remaining failure is explained above.

CI cannot be ticked before this run completes; it is the first execution of the new job and the point of the exercise.

## Review progress

- [ ] Code review
- [ ] Manual test

## Note for the reviewer

Two findings are worth knowing about.

First, the agent's include pattern had to be narrowed from `de.tum.cit.ase.ares.*` to `de.tum.cit.ase.ares.api.*`. The wider pattern also instrumented the integration test subjects, which are the very classes Ares analyses and enforces against. Three architecture tests of the ArchUnit/AspectJ combination then observed a different violation than the one they expect. This was confirmed by bisection: the identical 302-test selection passes without `-Pcoverage`, fails three tests with it, and passes again once the includes are narrowed. Since only `src/main/java` is reported on, and `api` is its only package, instrumenting test code never produced any coverage data to begin with.

That sensitivity itself deserves a separate look, independently of this pull request: it means a third-party agent attached alongside Ares can change which violation Ares reports.

Second, the thresholds are advisory on purpose. Merging only the unit slice with one AOP slice locally already gives 54.24% lines against the 55% minimum and 56.98% methods against the 65% minimum. The full aggregate will be higher, but nobody knows by how much. Once this run publishes real figures, the minima in `pom.xml` should be set to match reality, and dropping `-Djacoco.haltOnFailure=false` makes the gate binding.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -247,9 +247,8 @@ jobs:
     name: Coverage Report
     needs: [ unit-tests, tests, core-integration-tests ]
     # A failed slice still produced usable execution data, and the report is more useful
-    # then, not less, so this waits for the test jobs rather than requiring them to pass.
-    if: ${{ !cancelled() }}
-    runs-on: ubuntu-24.04
+    # than, not less, so this waits for the test jobs rather than requiring them to pass.
+    if: ${{ always() && !cancelled() }}
     timeout-minutes: 20
     permissions:
       contents: read
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -246,9 +246,13 @@ jobs:
   coverage-report:
     name: Coverage Report
     needs: [ unit-tests, tests, core-integration-tests ]
-    # A failed slice still produced usable execution data, and the report is more useful
-    # than, not less, so this waits for the test jobs rather than requiring them to pass.
-    if: ${{ always() && !cancelled() }}
+    # A failed slice still produced usable execution data, and the report is more useful in
+    # that case, not less, so this waits for the test jobs rather than requiring them to pass.
+    # `cancelled()` is a status check function, so naming it already suppresses the implicit
+    # `success()` on `needs`, and the job still runs after a failed slice. `always()` would
+    # be redundant here and GitHub discourages it.
+    if: ${{ !cancelled() }}
+    runs-on: ubuntu-24.04
     timeout-minutes: 20
     permissions:
       contents: read
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -120,12 +120,22 @@ jobs:
           cache: 'maven'
 
       - name: Unit tests
-        run: mvn test -Punit-core-tests -f pom.xml
+        run: mvn test -Punit-core-tests,coverage -f pom.xml
 
       # Architecture unit tests exercise the same execution-scoped WALA lifecycle
       # used by the integration matrices.
       - name: Architecture unit tests
-        run: mvn test -Punit-architecture-tests -f pom.xml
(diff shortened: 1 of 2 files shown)


ShudongCai (approved):
Code LGTM

PR #144, opened 2026-07-21
Stop the policy subsystem accepting values it cannot enforce

MarkusPaulsen:
## Summary

Closes the places in the policy subsystem where an omission, an ambiguity or an unchecked value
was silently accepted, replaces the exact policy-version check with a supported range applied by
both gates, rewrites the failure messages so a policy author is told what to write, and moves the
two validators to packages that describe what they are. This branch also carries a self-contained
refactor of the Java security-policy director, which no longer inspects the concrete type of its
collaborators to decide whether to rebind them to the discovered build configuration. Each
collaborator now exposes a `withBuildConfiguration` method and decides for itself, and the policy
documentation gains three new files under `docs/policy/`.

## Linked issues

None.

## 1. Problem

Not one defect but one recurring shape, found by reading the policy value objects closely. In a
security boundary the direction a mistake fails in matters more than its size, and in several
places the direction was outwards.

**The policy version could not express what a release supports.** `SecurityPolicy` carried a
single `CURRENT_POLICY_VERSION` and demanded exact equality with it, in two independent places:
the record's canonical constructor and the schema validator. A release understands a *set* of
format versions whose two ends move independently, and nothing forced those sites to stay in
step. Introducing a version 2 would have invalidated every existing version 1 policy file, or, if
only one site were updated, made the reader and the record disagree about the same file.

**A network permission built without a port permitted every port.** The builder held the port in
an `int`, so an unset port was `0`, and `0` is not neutral here: it means every port.
`builder().onTheHost("example.com").sendData(true).build()` therefore granted sending across the
whole host. Every other field in the package defaults to the restrictive value, the booleans to
`false` and the permission lists to empty. This one defaulted to the most permissive value it
could hold.

**Resource accesses built without an execution limit ran unbounded.** `createRestrictive()`
deliberately adds a limit, with a comment explaining that an empty list would let a
non-terminating submission run forever. The plain builder had the same empty default and no such
compensation, so the two disagreed on what "nothing configured" means.

**Two policy values were never checked at all.** Every other value had to match a pattern, on both
the parse path and the construction path. `ClassPermission` and `CommandPermission` accepted
anything non-blank, in either path, so a malformed class name or a command carrying a line break
was taken at face value.

**A rejected value was reported by quoting the regular expression it failed.** For a host that
meant several hundred characters of IPv6 alternation, which states what the expression is without
saying what to write instead. It was also the only message family in the policy package that
bypassed localisation, so it stayed English while its neighbours were translated.

**A command permission printed as though it had no argument constraint.** `CommandPermission`
overrode `toString()` to return the command alone, a remnant of serialising it as a bare string
through `@JsonValue`. The annotation was removed when that was found to drop arguments from
written policies; the override outlived it. A permission for `git` with arguments `[status]`
still printed as `git`, in exactly the diagnostics someone reads to find out why a command was
refused.

**Two classes sat in packages that misdescribed them.** `PolicyValueValidator` lived among the
records in `policySubComponents` although it is not a value object, leaving the reader importing
patterns out of a package of permissions. `SecurityPolicySchemaValidator` lived in `reader.yaml`
although nothing in it is YAML: it validates a Jackson node tree, which is the same tree whichever
format produced it. That placement invited a duplicate for the next format.

Finally, the documentation was wrong in ways a reader would act on: two Javadoc blocks in
`SupervisedCode` had run together onto one line so their tags never rendered, one named a
parameter that does not exist, `ResourceLimitsPermission` still called itself `TimeoutPermission`,
`ThreadPermission` documented only fully qualified class names though the validator also accepts a
wildcard and six concurrency tokens, and nothing in the package documented what it throws.

The value-hardening root cause sits in the policy layer, in the value objects and their reader;
the separate director refactor described in §6 additionally touches `SecurityPolicyJavaDirector`
and its Java collaborators in `securitytest.java`. Neither part changes enforcement, AOP or
architecture behaviour.

- An instructor assembling a policy in code can no longer grant a whole host by forgetting a
  port, or an unbounded run by forgetting a limit. Both now have to be written down.
- A rejected value now says what the field accepts and quotes the offending value:
  `onTheHost must be a DNS name, an IPv4 or IPv6 address, localhost, or * for every host, but was
  'not a host'`, in English and German, instead of the raw pattern in English only.
- A malformed class name or a command containing a control character is refused when the policy
  is read, rather than accepted and then matching nothing during enforcement.
- A refused command is reported with its arguments, so a denial that hinges on the arguments can
  be understood from the message rather than from the source.
- The Javadoc now describes the wildcard and the six concurrency tokens `ThreadPermission`
  accepts, the placeholders a file path may use, and what every entry point throws.
- Once a version 2 of the policy format exists, an existing version 1 file keeps loading. That
  compatibility is prepared here, not claimed here: the range is `[1, 1]`, so no policy file
  changes behaviour.

- One private helper is the single source of truth for the version invariant, so the constructor
  and the builder cannot drift apart in either the condition or the message.
- The schema validator expresses the same invariants as the records, for the version and now for
  commands too, so raising `MAXIMUM_POLICY_VERSION` is a one-line change rather than a hunt for
  hard-coded equality.
- The remaining defaults across the package all fail closed, so "unset" has one meaning.
- The message wording lives beside the pattern it explains, and an unmapped pattern still
  produces the old form rather than no message at all.
- `SecurityPolicySchemaValidator` now records why it is not an interface with one implementation
  per format, so the next person does not build that seam where nothing varies.
- The package produces zero Javadoc warnings, down from one before this branch. The previous
  `{@value}` tag pointed at the removed `CURRENT_POLICY_VERSION`, which fails
  `maven-javadoc-plugin`, configured here with `failOnError`.
- The policy subsystem gains three new documentation files under `docs/policy/`:
  `SecurityPolicyModel.puml` (the policy value-object model), `SecurityPolicyReaderAndDirectorClassDiagram.puml`
  (the reader-and-director structure, including the new `ProgrammingLanguageConfiguration` enum and
  its dependency on `SecurityPolicyDirector`), and `UMLClassDiagramRules.txt` (the conventions those
  diagrams follow).

## 4. Testing manual

**Prerequisites**

1. JDK 21 and Maven, as for any Ares 2 build. No echo server and no exercise repository are
   needed: everything below is reachable from this repository alone.
2. Check out this branch.

**Steps**

1. Run the policy suite:
   `mvn test -Dtest='de.tum.cit.ase.ares.api.policy.**' -DfailIfNoSpecifiedTests=false`
2. Confirm the Javadoc builds, which the state before this branch did not:
   `mvn javadoc:javadoc`
3. Take any policy file under `src/test/resources/.../securitypolicies/`, copy it, and set
   `thisPolicyFileCompliesToThePolicyVersion` to `0`, then `2`, then `4294967297`. Load each with
   `SecurityPolicyYAMLReader`.
4. In the same copy, set a command permission to `executeTheCommand: "ec\nho"`, and separately a
   file permission host to `onTheHost: "not a host"`. Load each.
5. In a scratch class or `jshell`, evaluate each of:
   - `NetworkPermission.builder().onTheHost("example.com").sendData(true).build()`
   - `ResourceAccesses.builder().build().regardingTimeouts()`
   - `ResourceAccesses.builder().regardingTimeouts(List.of()).build().regardingTimeouts()`
   - `new CommandPermission("git", List.of("status")).toString()`
   - `new ClassPermission("not a class name")`

**Expected result**

1. 335 tests pass.
2. `BUILD SUCCESS`, with no warning for any file under `api/policy`.
3. Each file is rejected with a `SecurityException`. The cause reads `must be between 1 and 1
   (inclusive), but was 0` and `... but was 2` for the first two, and `must fit into an int` for
   the third, which is integral but truncates to `1`. The unmodified file, declaring version 1,
   still loads.
4. Both are rejected, the first for the control character and the second with the host message
   quoted above, in the language of the active locale.
5. In order: a `NullPointerException` reading `onThePort must not be null`;
   `[ResourceLimitsPermission[timeout=10000]]`; `[]`, showing the opt-out is still reachable
   explicitly; `CommandPermission[executeTheCommand=git, withTheseArguments=[status]]`; and an
   `IllegalArgumentException` naming what a class name must look like.

The set of accepted policy *files* is unchanged for every fixture in this repository, and all 419
tests covering them pass. The behavioural changes are reachable through the programmatic builders,
which no policy file goes through, and through two values that were previously unchecked.
Specifically:

- A policy declaring any version other than 1 is still rejected, at both gates.
- Permitting every port is still expressible, by writing `0` deliberately. What is no longer
  expressible is permitting every port by accident.
- An unbounded execution time is still expressible, by passing an empty list. What is no longer
  expressible is an unbounded run by omission.
- A command may still be a bare name, a path, or `*`; an argument may still be empty or `*`. What
  is refused is a control character in either, and whitespace around a command.
- A class permission may still name a package prefix, which is how six of the twenty-two
  essential classes are declared and how prefix matching is meant to work.
- No permission became easier to satisfy: the diff adds rejections and removes none.

**Modes exercised**

- [ ] ArchUnit + AspectJ
- [ ] WALA + AspectJ
- [ ] WALA + instrumentation

Instruction, Branch, Line and Method coverage for every production class changed in this PR, read from the aggregated JaCoCo report produced by the "Coverage Report" job of this PR's Maven workflow run. Nested classes (for example a record's `Builder`) are listed as their own rows. `n/a` means the counter has no members (a record builder with no branches, `AresConstants`' branch counter, or the `LanguageNameRules` interface).

| `SecurityPolicy` | 100.0% (60/60) | 100.0% (4/4) | 100.0% (12/12) | 100.0% (4/4) |
| `SecurityPolicy.Builder` | 100.0% (31/31) | n/a | 100.0% (10/10) | 100.0% (4/4) |
| `SupervisedCode` | 100.0% (87/87) | 100.0% (6/6) | 100.0% (24/24) | 100.0% (3/3) |
| `SupervisedCode.Builder` | 100.0% (63/63) | n/a | 100.0% (23/23) | 100.0% (7/7) |
| `ResourceAccesses` | 100.0% (96/96) | n/a | 100.0% (19/19) | 100.0% (3/3) |
| `ResourceAccesses.Builder` | 100.0% (135/135) | n/a | 100.0% (34/34) | 100.0% (8/8) |
| `FilePermission` | 100.0% (51/51) | n/a | 100.0% (9/9) | 100.0% (3/3) |
| `FilePermission.Builder` | 100.0% (55/55) | n/a | 100.0% (17/17) | 100.0% (8/8) |
| `NetworkPermission` | 100.0% (59/59) | 100.0% (4/4) | 100.0% (9/9) | 100.0% (3/3) |
| `NetworkPermission.Builder` | 100.0% (50/50) | n/a | 100.0% (14/14) | 100.0% (7/7) |
| `ThreadPermission` | 100.0% (37/37) | 100.0% (2/2) | 100.0% (8/8) | 100.0% (3/3) |
| `ThreadPermission.Builder` | 100.0% (27/27) | n/a | 100.0% (7/7) | 100.0% (4/4) |
| `PackagePermission` | 100.0% (22/22) | n/a | 100.0% (8/8) | 100.0% (3/3) |
| `PackagePermission.Builder` | 100.0% (20/20) | n/a | 100.0% (6/6) | 100.0% (3/3) |
| `ResourceLimitsPermission` | 100.0% (27/27) | 100.0% (2/2) | 100.0% (6/6) | 100.0% (3/3) |
| `ResourceLimitsPermission.Builder` | 100.0% (14/14) | n/a | 100.0% (4/4) | 100.0% (3/3) |
| `ClassPermission` | 100.0% (25/25) | 100.0% (2/2) | 100.0% (6/6) | 100.0% (2/2) |
| `ClassPermission.Builder` | 100.0% (20/20) | n/a | 100.0% (4/4) | 100.0% (3/3) |
| `CommandPermission` | 100.0% (84/84) | 100.0% (4/4) | 100.0% (19/19) | 100.0% (4/4) |
| `CommandPermission.Builder` | 100.0% (36/36) | n/a | 100.0% (9/9) | 100.0% (4/4) |
| `ProgrammingLanguageConfiguration` | 100.0% (92/92) | n/a | 100.0% (19/19) | 100.0% (2/2) |
| `JavaNameRules` | 100.0% (111/111) | 100.0% (10/10) | 100.0% (23/23) | 100.0% (9/9) |
| `LanguageNameRules` | n/a | n/a | n/a | n/a |
| `PolicyValueValidator` | 100.0% (96/96) | 100.0% (8/8) | 100.0% (20/20) | 100.0% (4/4) |
| `SecurityPolicySchemaValidator` | 89.7% (627/699) | 93.8% (105/112) | 85.4% (135/158) | 100.0% (15/15) |
| `SecurityPolicyReader` | 100.0% (90/90) | 100.0% (8/8) | 100.0% (21/21) | 100.0% (4/4) |
| `SecurityPolicyYAMLReader` | 100.0% (158/158) | 100.0% (6/6) | 100.0% (32/32) | 100.0% (4/4) |
| `SecurityPolicyYAMLReader.Builder` | 100.0% (30/30) | n/a | 100.0% (7/7) | 100.0% (4/4) |
| `SecurityPolicyReaderAndDirector` | 100.0% (108/108) | 100.0% (4/4) | 100.0% (30/30) | 100.0% (8/8) |
| `SecurityPolicyReaderAndDirector.Builder` | 100.0% (37/37) | n/a | 100.0% (9/9) | 100.0% (5/5) |
| `SecurityPolicyDirector` | 100.0% (86/86) | 100.0% (2/2) | 100.0% (21/21) | 100.0% (3/3) |
| `SecurityPolicyJavaDirector` | 100.0% (275/275) | 100.0% (16/16) | 100.0% (57/57) | 100.0% (7/7) |
| `SecurityPolicyJavaDirector.Builder` | 100.0% (98/98) | n/a | 100.0% (25/25) | 100.0% (9/9) |
| `AresConstants` | 100.0% (14/14) | n/a | 100.0% (2/2) | 100.0% (1/1) |
| `JavaCreator` | 100.0% (504/504) | 100.0% (31/31) | 100.0% (108/108) | 100.0% (29/29) |
| `JavaWriter` | 97.8% (438/448) | 90.0% (18/20) | 100.0% (98/98) | 100.0% (13/13) |
| `JavaProjectScanner` | 96.2% (718/746) | 94.1% (96/102) | 95.7% (111/116) | 96.9% (31/32) |
| `JavaProgrammingExerciseProjectScanner` | 100.0% (26/26) | 100.0% (2/2) | 100.0% (10/10) | 100.0% (5/5) |
| **Total (changed classes)** | **97.6% (4507/4617)** | **95.7% (330/345)** | **97.2% (976/1004)** | **99.6% (235/236)** |

Every measurable counter reaches 100% except the cells below, which cannot be covered without deleting defensive code or depend on the runtime environment (this PR keeps all defensive code and makes no production changes just to move a number):

- **Unreachable by design.** The `node == null` / `commands == null` guards in `SecurityPolicySchemaValidator`'s `requireText`, `requireIntegral`, `requireBooleans`, `requireTextArray`, `requireTextArrayAllowingBlankEntries`, `validateObjectArray`, `validateCommandArray` and `requireObject` — `requireObject` already rejects any missing or null required field upstream, so no input reaches them (their `!isArray` / `!isObject` / value-level sides are covered). `JavaWriter.createLocalisationFiles`'s `testFolderPath.toString().isEmpty()` and `getRoot() == null`, which cannot occur because `BuildToolConfiguration.canonicalise` returns an absolute path. `JavaProjectScanner.parse`'s `orElseThrow` supplier, which JavaParser never invokes for a successful parse.
- **Not deterministically exercisable under the Maven CI checkout.** `JavaProjectScanner.testRoots`'s no-`buildConfiguration` Gradle/Maven fallback (the unit tests always supply a `BuildToolConfiguration`, so the working-directory-dependent fallback is never taken) and `javaFiles`'s `!isReadable` branch (filesystem-permission dependent).

## 6. Follow-up refactor: polymorphic collaborator configuration

`SecurityPolicyJavaDirector` chose whether to rebind a collaborator to the discovered
`BuildToolConfiguration` by inspecting its concrete type (`getClass() == JavaCreator.class`,
`instanceof JavaProgrammingExerciseProjectScanner`). That decision now lives in a
`withBuildConfiguration(BuildToolConfiguration)` method on `JavaCreator`, `JavaWriter` and
`JavaProjectScanner`, overridden in `JavaProgrammingExerciseProjectScanner`: an exact
framework-default type is rebound to the discovered configuration, and any subclass is returned
unchanged so a custom injected collaborator is never silently replaced. Scanner subclasses are now
preserved as well, consistent with the creator and writer, and this compatibility change is
documented on the method. Direct tests exercise exact-type rebinding, subclass preservation and
null rejection, and the reader-and-director class diagram (`docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml`)
adds the `ProgrammingLanguageConfiguration` enum and its dependency on `SecurityPolicyDirector`.

Nothing here changes a policy file, an exercise repository or the generated security test code, so
no instructor has to touch an existing exercise. Six changes affect code that constructs or
inspects policy objects programmatically, and they should be weighed rather than waved through.

1. **`SecurityPolicy.CURRENT_POLICY_VERSION` is removed**, replaced by `MINIMUM_POLICY_VERSION`
   and `MAXIMUM_POLICY_VERSION`. This constant is not new and unreleased: it shipped in **v2.1.0**,
   tagged 2026-07-21, so it is part of a released artefact. Migration is a rename to
   `MAXIMUM_POLICY_VERSION`, which carries the same value and meaning. If that is judged too
   aggressive, keeping the old name as a `@Deprecated` alias for one release is the alternative;
   this branch takes the clean removal deliberately, for review.
2. **`PolicyValueValidator` moves** from `…policy.policySubComponents` to `…policy`. It is a
   public class that also shipped in v2.1.0, so any external import of it stops compiling.
   Migration is the changed import. It was **not** moved to `api.util` on purpose: nine of its
   constants and half its methods are policy vocabulary, encoding the implicit-thread-op tokens,
   the recognised path placeholders and the eight configuration names, and all ten of its
   consumers live inside `api.policy`. Moving it to `util` would put policy semantics in a
   general utility package.
3. **`NetworkPermission.Builder.build()` now throws when the port was never set**, instead of
   defaulting to `0` and thereby permitting every port. Callers relying on the default must pass
   `0` explicitly. Every call site in this repository already sets the port.
4. **`ResourceAccesses.Builder` now starts with the restrictive execution limit** instead of an
   empty list. A caller who wants no limit passes an empty list explicitly. This tightens the
   result of any builder that omitted it.
5. **`CommandPermission.toString()` now returns the record's own representation**, including the
   arguments, rather than the bare command. Nothing in this repository parses it; anything
   downstream that did should read `executeTheCommand()` instead.
6. **`ClassPermission` and `CommandPermission` now reject values they previously accepted**: a
   class name that is not a dot-separated Java name, and a command or argument containing a
   control character, or a command with surrounding whitespace. Everything the repository's
   fixtures and the twenty-two essential classes actually use was checked against the patterns
   before adopting them.
7. **`JavaCreator`, `JavaWriter` and `JavaProjectScanner` gain a public `withBuildConfiguration`
   method**, and custom *scanner* subclasses are now preserved rather than replaced by a
   reconfigured plain scanner. A custom scanner subclass that needs the discovered configuration
   must now override this method. No production caller is affected: the framework defaults behave
   exactly as before.

One further change is deliberately **not** in that list. `SecurityPolicySchemaValidator` moves
from `…reader.yaml` to `…reader` and becomes public, but it was package-private before, so no
external code could reference it: the move is invisible outside the artefact and the new
visibility is an addition, not a break. It had to become public because its caller now sits in a
sub-package and can no longer see a package-private type; its Javadoc records that it is internal
to the reader.

## Checklist

- [x] Documentation (`docs/`, `README.adoc`, Javadoc) was updated where the change is user-facing.

## Review progress

- [ ] Code review
- [ ] Manual test


src/main/java/de/tum/cit/ase/ares/api/policy/reader/yaml/SecurityPolicySchemaValidator.java
```diff
@@ -42,9 +42,11 @@ private SecurityPolicySchemaValidator() {
...
 		requireIntegral(root, "thisPolicyFileCompliesToThePolicyVersion", "$");
-		if (!root.get("thisPolicyFileCompliesToThePolicyVersion").canConvertToInt() || root
-				.get("thisPolicyFileCompliesToThePolicyVersion").intValue() != SecurityPolicy.CURRENT_POLICY_VERSION) {
-			fail("$.thisPolicyFileCompliesToThePolicyVersion must be exactly " + SecurityPolicy.CURRENT_POLICY_VERSION);
+		JsonNode policyVersion = root.get("thisPolicyFileCompliesToThePolicyVersion");
+		if (!policyVersion.canConvertToInt() || policyVersion.intValue() < SecurityPolicy.MINIMUM_POLICY_VERSION
+				|| policyVersion.intValue() > SecurityPolicy.MAXIMUM_POLICY_VERSION) {
+			fail("$.thisPolicyFileCompliesToThePolicyVersion must be between " + SecurityPolicy.MINIMUM_POLICY_VERSION
+					+ " and " + SecurityPolicy.MAXIMUM_POLICY_VERSION + " (inclusive)");
 		}
```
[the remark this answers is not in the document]
MarkusPaulsen: Applied in 0c7f2d27, though not by caching alone.

Hoisting the value would have removed the repetition while keeping the real
problem: one condition, and therefore one message, for two unrelated faults. A
version that does not fit into an `int` was reported as lying outside the
supported range, which is true but tells the policy author nothing, since
`4294967297` is not out of range in any sense they would recognise. It simply is
not an `int`.

The guard is now split, so each fault is named on its own terms, and the range
message quotes the offending value as the record's message already did:

| Declared version | Before | Now |
| --- | --- | --- |
| `4294967297` | must be between 1 and 1 (inclusive) | must fit into an int |
| `2` | must be between 1 and 1 (inclusive) | must be between 1 and 1 (inclusive), but was 2 |

Both were verified by loading such a policy through `SecurityPolicyYAMLReader`
and reading the cause of the resulting `SecurityException`, not only from the
tests. Reading the converted value once instead of twice falls out of the split.

All six branch outcomes of the two conditions are covered.

src/main/java/de/tum/cit/ase/ares/api/policy/SecurityPolicy.java
```diff
@@ -14,55 +14,108 @@
...
+	/**
+	 * The latest policy-format version supported by this Ares release.
+	 * <p>
+	 * This is the version assumed for policies that are built programmatically
+	 * without declaring one, as such policies are by construction expressed in the
+	 * newest format this release understands.
+	 *
+	 * @since 2.1.0
+	 */
+	public static final int MAXIMUM_POLICY_VERSION = 1;
```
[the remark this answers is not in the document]
MarkusPaulsen: Applied in 0c7f2d27.

One point worth recording for whoever reads this next, because it is not obvious
and it was the crux of the decision. I had argued against the guard on the
grounds that an unreachable `true` branch would leave a permanently uncovered
branch in a class that is otherwise at 100%. That reasoning was wrong, and
checking the bytecode rather than assuming settled it:

```
javap -p target/classes/.../SecurityPolicy.class   ->   no <clinit> in the class
```

Both bounds are compile-time constants, so `1 > 1` folds to `false` and javac
eliminates the block entirely under the conditional-compilation rule. The guard
is zero bytecode while the range is sound, and materialises only once an edit
makes the two ends inconsistent, which is exactly when it is wanted. JaCoCo
confirms `SecurityPolicy` is unchanged at 13/13 lines and 4/4 branches.

For the record, the premise of the suggestion still overstates the risk: an
inconsistent range is not silent, because every policy would be rejected and the
whole policy suite would fail at once. The guard was worth taking regardless,
since it names the cause and costs nothing.

SedaOran (dismissed):
Overall this looks great to me - happy to approve. 🙂 

A couple of small thoughts, purely as suggestions - nothing blocking:

- I was thinking it might be worth mentioning two of these in the release notes / CHANGELOG, since they're intentional but slightly behaviour-changing for anyone using Ares as a library and building policies in code: NetworkPermission.Builder.build() now requires a port instead of defaulting to 0 (every port), and ResourceAccesses.Builder now starts with the restrictive 10s limit instead of an unbounded default. Both feel like the right call -  I just thought a short "library-consumer breaking" note could save a downstream consumer some head-scratching.
- On the serialises/serializes alignment (which I think is actually a nice fix, since the switch case never matched the localized rule name before) - I noticed the rule name still comes from the localized message, so under a non-English locale the mapping would fall through to the default. That's pre-existing and clearly out of scope here, so maybe it's something we could spin off into a separate follow-up issue whenever it feels worth it, rather than touching it in this PR.
Thanks for the thorough work on this - the fail-closed direction throughout is really reassuring on a security boundary.

src/main/java/de/tum/cit/ase/ares/api/policy/PolicyValueValidator.java
```diff
@@ -1,5 +1,6 @@
-package de.tum.cit.ase.ares.api.policy.policySubComponents;
+package de.tum.cit.ase.ares.api.policy;
 
+import java.util.Map;
```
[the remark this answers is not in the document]
MarkusPaulsen: Not adding the shim. `PolicyValueValidator` is an internal validation helper: every reference lives inside Ares's own `policy` packages (`SecurityPolicySchemaValidator`, `CommandPermission`, `FilePermission`, `NetworkPermission`, `JavaNameRules`), and 2.1.0 has not been released, so the pre-move `policySubComponents` location was never part of a published artifact. There are no downstream imports to preserve, so a deprecated forwarding class would be permanent dead weight.

src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/ThreadPermission.java
```diff
@@ -31,22 +42,30 @@ public record ThreadPermission(int createTheFollowingNumberOfThreads, @Nonnull S
...
 	public ThreadPermission {
 		Objects.requireNonNull(ofThisClass, "Thread class must not be null");
 		if (createTheFollowingNumberOfThreads < 0) {
 			throw new IllegalArgumentException(Messages.localized("policy.permission.thread.count.negative"));
 		}
-		PolicyValueValidator.requireMatch("ofThisClass", ofThisClass, PolicyValueValidator.THREAD_CLASS_PATTERN);
+		// Only the universal invariants are checked here: this record does not carry
+		// the language, so whether ofThisClass is a valid thread construct for the
+		// supervised code's language is validated by the schema validator on the
+		// policy-file path.
```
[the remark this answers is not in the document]
MarkusPaulsen: Keeping the current design. `ThreadPermission` deliberately checks only the universal invariants (non-null selector, non-negative count) and defers whether `ofThisClass` is a valid thread construct to the schema validator, because the record does not carry the language it would need to make that judgement. This split is stated in the record's Javadoc and mirrors how the other language-dependent names (packages, classes, thread constructs) are validated on the policy-file path. Moving it into `SupervisedCode` would be a broader architectural change than this PR's scope and is a conscious design decision rather than an oversight.

src/main/java/de/tum/cit/ase/ares/api/policy/reader/SecurityPolicySchemaValidator.java
```diff
@@ -115,11 +145,22 @@ private static void validateCommandArray(JsonNode commands) throws MismatchedInp
 				if (command.textValue().isBlank()) {
 					fail("A bare command permission must not be blank");
 				}
+				if (!PolicyValueValidator.matches(command.textValue(), PolicyValueValidator.COMMAND_PATTERN)) {
+					fail("A bare command permission must not contain control characters");
+				}
```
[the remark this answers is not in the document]
MarkusPaulsen: This is addressed by the current shape. The bare-command check now runs through `requirePattern`, whose message points at the expected command format rather than naming only control characters, and the argument check's "must not contain control characters" is accurate for `COMMAND_ARGUMENT_PATTERN` (which rejects controls and line separators but permits surrounding whitespace). Resolving as outdated.

docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml
```diff
@@ -0,0 +1,313 @@
...
+}
+
+note right of SecurityPolicyJavaDirector
+      DEFAULT_ESSENTIAL_PACKAGES_PATH: nonnull
+      - allowed: EssentialPackages.yaml source path
+      DEFAULT_ESSENTIAL_CLASSES_PATH: nonnull
+      - allowed: EssentialClasses.yaml source path
+end note
+
+SecurityPolicyReaderAndDirector "1" -[norank]-> "0..1" SecurityPolicyDirector : - securityPolicyDirector: SecurityPolicyDirector
```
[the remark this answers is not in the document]
MarkusPaulsen: Leaving this edge rank-neutral on purpose. `SecurityPolicyDirector` is drawn above `SecurityPolicyReaderAndDirector`, so this association points upward, and the `[norank]` modifier keeps it from distorting the ranking that the hidden directional links in this diagram pin by hand. Rule 21 lists only `left`, `right` and `down`, so there is no explicit keyword that expresses an upward association without misrouting it; forcing one of the three would fight the intended layout. The sibling dependency edges (lines 268-270) use the same rank-neutral style for the same reason.

MarkusPaulsen:
@coderabbitai review

src/test/java/de/tum/cit/ase/ares/api/securitytest/java/projectScanner/JavaProjectScannerAstTest.java
```diff
@@ -95,6 +99,80 @@ void reportsAConfiguredNonDirectorySourceRoot() throws IOException {
...
+				class LegacyCase extends TestCase { public void testSomething() {} }
+				@Test class ClassLevelAnnotated {}
+				class PlainHelper extends Object { void doWork() {} }
+				""");
+		JavaProjectScanner scanner = new JavaProjectScanner(configuration(production, tests));
+		// Main is preferred over Application, which is preferred over any other name.
+		assertEquals("Main", scanner.scanForMainClassInPackage());
+		// A JUnit 3 class extending TestCase and a class with a class-level test
+		// annotation are both recognised, while a plain class that does neither is not.
+		assertArrayEquals(new String[] { "pref.ClassLevelAnnotated", "pref.LegacyCase" }, scanner.scanForTestClasses());
```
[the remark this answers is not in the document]
MarkusPaulsen: Half applied, half declined, and the halves point in opposite directions.

**Applied: the class-level annotations.** You are right that Jupiter's `@Test` targets only
`ANNOTATION_TYPE` and `METHOD`, so `@Test class Foo {}` is not Java a compiler would accept and
the fixture documented a shape that cannot reach the scanner from a real project. What makes the
type-level branch worth having is the composed case: `@Test` on `ANNOTATION_TYPE` is legal, so a
project-defined annotation may carry `@Test` and declare `@Target(TYPE)` for itself, and
`@ThatAnnotation class Foo {}` then compiles. That is the path
`AnnotationResolver.trustComposedAnnotations` exists for. Both fixtures now use it:

- `LegacyTests.java` declares `@Test @Target(ElementType.TYPE) @interface ClassLevelTest {}` and
  applies it, resolved through the in-unit declaration.
- `TypeLevelTest.java` declares the same shape fully qualified, used from its own package by
  `TypeAnnotated` and through a single-type import by the default-package `DefaultPackageCase`,
  so the package lookup and the import path are both exercised, which the two bare
  `@org.junit.jupiter.api.Test` uses did not distinguish.

The assertions are unchanged, so the recognition behaviour is pinned exactly as before, now from
fixtures that compile. The annotation declarations are correctly not reported as test classes,
which `scanForTestClasses` skips explicitly.

**Declined: the missing `import junit.framework.TestCase`.** That omission is the test, not an
oversight. `extendsTestCase` reads

```java
return candidates.isEmpty() || candidates.equals(Set.of(JUNIT_THREE_TEST_CASE));
```

so an unresolvable bare `TestCase` is accepted deliberately: refusing it would supervise a JUnit 3
test class rather than exempt it, and the scan has no compiler to ask. The imported spelling is
already covered by the fixture in
`prefersMainThenApplicationAndRecognisesJUnitThreeTestCases`, which does import it. The two
fixtures together cover both sides of that condition; adding the import here would collapse them
onto one side and leave `candidates.isEmpty()` untested. I have added a comment saying so, since
the next reader will otherwise reach for the same fix.

Verified with `mvn -o test -Dtest=JavaProjectScannerAstTest`: 29 tests, 0 failures.

MarkusPaulsen:
@Claudia-Anthropica All three findings from the review of 12 August are addressed
in `8d954a9e`. Replying here rather than in a thread, since the review carried no
inline comments.

**1. Version-1 command policies and the removed `fromJson`/`fromString`.** Correct
on both counts, and the second was the part I had not weighed properly. Removing
the bare scalar form was a deliberate narrowing of a format whose version number
did not move: `MINIMUM_POLICY_VERSION` and `MAXIMUM_POLICY_VERSION` both still say
1, so a file that loaded under v2.1.0 has to keep loading. Both methods return as
`@Deprecated(forRemoval = true)`, and with them the scalar form, keeping the
meaning it had, the command with an **empty** argument list. That is pinned by
`SecurityPolicyStrictSchemaTest.stillLoadsTheDeprecatedBareScalarCommandForm`, so
it cannot quietly widen to "any arguments" later, and the schema gate holds the
scalar to `COMMAND_PATTERN` exactly as it holds the mapping form. Both go with the
next policy-format version.

Restoring the creator turned out to matter for a second reason: a `@JsonCreator`
binds the whole node itself, so Jackson's own unknown-property handling never runs
for this record. The restored creator therefore enforces the released two-field
contract, and a new test covers a misspelt field, which would otherwise have bound
silently through a bare mapper and dropped the declared arguments.

**2. The abstract three-argument `createTestCases`.** Correct, and the
`AbstractMethodError` for an already-compiled subclass is the worse half of it. It
is concrete again, taking the middle course you proposed rather than simply
reverting: the reason it was made abstract stands, because the default it shipped
with forwarded unconditionally and so analysed the whole project when the caller
had asked for one directory. The inherited default now forwards only the empty
path, which is how an unscoped analysis is written throughout the policy layer,
and throws `UnsupportedOperationException` for any real scope. Three tests cover
concreteness, forwarding and refusal.

**3. `createRestrictive()` at 3,000 ms.** The change is intended, the description
was not. It is now breaking change 8, which says what it tightens and how to
restore the previous budget, and the expected result in §4 no longer claims
10,000. The test counts in §4 were stale too and now read 456, verified locally.

Not changed, and I want to flag the reasoning rather than leave it implicit:
`CommandPermission.toString()` stays the record representation. That is also a
behavioural break against v2.1.x, but it is a deliberate one, listed as breaking
change 5, because the old override returned the bare command and so dropped the
arguments from exactly the diagnostics someone reads to find out why a command was
refused.

Verified locally with JDK 17: 456 policy tests and 29 scanner tests, no failures.

src/main/java/de/tum/cit/ase/ares/api/policy/SecurityPolicy.java
```diff
@@ -14,55 +16,94 @@
...
 @SuppressWarnings("unused")
 public record SecurityPolicy(int thisPolicyFileCompliesToThePolicyVersion,
 		@Nonnull SupervisedCode regardingTheSupervisedCode) {
 
-	/** The policy-format version supported by this Ares release. */
-	public static final int CURRENT_POLICY_VERSION = 1;
-
 	/**
-	 * Constructs a SecurityPolicy instance with validated supervised code.
+	 * Constructs a SecurityPolicy instance with a validated policy-format version
```
Claudia-Anthropica, line 68: @MarkusPaulsen [medium] This refactor removes `SecurityPolicy.CURRENT_POLICY_VERSION`, a public constant shipped in v2.1.0 through v2.1.2, so downstream sources using it no longer compile. Keep it as a deprecated alias of `AresConstants.MAXIMUM_POLICY_VERSION` for at least one compatibility cycle.

In `src/main/java/de/tum/cit/ase/ares/api/policy/SecurityPolicy.java`, removing `CURRENT_POLICY_VERSION` breaks source compatibility for consumers of the v2.1.x public API. Restore it as a deprecated alias of `AresConstants.MAXIMUM_POLICY_VERSION` for at least one compatibility cycle.

</details>

docs/policy/SecurityPolicyModel.puml
```diff
@@ -0,0 +1,272 @@
...
+    + timeout: long
+    -- Factories --
+    + {static} createRestrictive(): ResourceLimitsPermission
+    + {static} builder(): ResourceLimitsPermission.Builder
+}
+
+note right of ResourceLimitsPermission
+  timeout: nonnull
+  - limits: > 0 ms
+  - default: 10,000 ms
```
Claudia-Anthropica: @MarkusPaulsen [low] `ResourceLimitsPermission.createRestrictive()` now returns 3,000 ms and the manual was updated accordingly, but this new domain-model diagram still advertises 10,000 ms. Change the diagram default to 3,000 ms so the documented policy contract matches the implementation.

In `docs/policy/SecurityPolicyModel.puml`, the restrictive timeout is documented as 10,000 ms even though `ResourceLimitsPermission.createRestrictive()` returns 3,000 ms. Update the diagram default to 3,000 ms.

</details>

src/main/java/de/tum/cit/ase/ares/api/policy/SecurityPolicy.java
MarkusPaulsen: Applied in `159d6f90`, and the finding is consistent with the two compatibility
corrections in `8d954a9e`, so it should have been caught in the same pass.

`CURRENT_POLICY_VERSION` returns as `@Deprecated(forRemoval = true)`, aliasing
`AresConstants.MAXIMUM_POLICY_VERSION`, which carries the value and the meaning it
had: it meant the version this release writes, which is what
`SecurityPolicy.Builder` still defaults to. The maximum, not the minimum, is
therefore the right target.

Its Javadoc records one caveat the plain rename hides. The removed constant was
compared for *equality* at both gates, and that is precisely what the range
replaces. A downstream caller that migrates the name but keeps `==` will reject a
version 1 policy the moment version 2 exists, which is the failure this branch set
out to prevent. The deprecation text says so rather than leaving the rename to look
free.

Pinned by `AresConstantsTest.keepsTheReleasedVersionConstantAsAnAliasOfTheMaximum`.
459 policy tests pass, and `mvn clean package` is green, which now matters here:
the SpotBugs and Javadoc gates run in the `package` phase, and the first push of
`8d954a9e` turned three CI jobs red on a single SpotBugs finding that no test run
would have shown.

Breaking change 1 in the description has been rewritten accordingly; it previously
announced the removal as deliberate.

docs/policy/SecurityPolicyModel.puml
MarkusPaulsen: Correct, and left open on purpose for now.

`docs/policy/SecurityPolicyModel.puml` is being edited in the working copy at the
moment, so a one-line change to it from this side would either be overwritten or
drag unrelated work in progress into a commit. The committed line 206 does read
`- default: 10,000 ms` and has to become `3,000 ms` to match
`ResourceLimitsPermission.createRestrictive()`, which is now breaking change 8 in
the description.

Worth noting for whoever makes the edit: the same diagram's `ResourceAccesses` note
is affected by the same commit, since `ResourceAccesses.Builder` now starts from
`createRestrictive()` rather than from an empty list, so "default" there means the
same 3,000 ms rather than "no limit". Changing only the number on line 206 would
leave the diagram consistent about the value and still wrong about where it applies.

This thread stays open until that lands.

MarkusPaulsen: Correction to my previous reply: the `ResourceAccesses` note needs nothing. Lines
92 to 93 read `regardingTimeouts: nonnull / - allowed: immutable list with one
restrictive default`, which names no number and already describes the builder
change correctly.

The edit is the single number on line 206, `10,000 ms` to `3,000 ms`. Nothing
else in the diagram carries the value.

docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml
```diff
@@ -0,0 +1,358 @@
...
+    # creator: Creator
+    # writer: Writer
+    # executer: Executer
+    # essentialDataReader: EssentialDataReader
+    # projectScanner: ProjectScanner
+    # essentialPackagesPath: Path
+    # essentialClassesPath: Path
+    + SecurityPolicyDirector(creator: Creator, writer: Writer, executer: Executer, essentialDataReader: EssentialDataReader, ...)
+    + {abstract} createTestCases(securityPolicy: SecurityPolicy, projectRootPath: Path): TestCaseAbstractFactoryAndBuilder
+    + {abstract} createTestCases(securityPolicy: SecurityPolicy, projectRootPath: Path, withinPath: Path): TestCaseAbstractFactoryAndBuilder
```
Claudia-Anthropica: @MarkusPaulsen [low] This diagram marks the three-argument `createTestCases` overload abstract, but `SecurityPolicyDirector` intentionally keeps it concrete to avoid breaking subclasses compiled against v2.1.x. Remove `{abstract}` from this overload so the diagram matches the compatibility implementation.

In `docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml`, the three-argument `SecurityPolicyDirector.createTestCases` overload is shown as abstract even though the implementation keeps it concrete for binary compatibility. Remove `{abstract}` from that overload declaration.

</details>

MarkusPaulsen: Correct, and left open like the `SecurityPolicyModel.puml` thread, for the same
reason: both diagrams are being edited in the working copy right now, so a change
from this side would either be overwritten or pull unrelated work in progress into
a commit.

The edit is line 277 of the committed file, dropping `{abstract}` from the
three-argument overload. The two-argument overload on line 276 keeps it and is
still genuinely abstract.

One thing the diagram would then understate. Concrete here does not mean the
inherited default does the work: it forwards only `Path.of("")`, the unscoped
convention, and throws `UnsupportedOperationException` for any real `withinPath`
rather than silently analysing the whole project, which is what the released
default did. If the notation carries such conditions elsewhere, a note saying so
is worth more than the removed marker, since "concrete" on its own reads as "a
subclass need not override it", and for a director that wants a scope it must.

src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/SupervisedCode.java
```diff
@@ -46,38 +47,61 @@ public record SupervisedCode(
...
+	 *                                  one of the test class names does not have
+	 *                                  the shape Java requires of it.
 	 */
 	public SupervisedCode {
 		Objects.requireNonNull(theFollowingProgrammingLanguageConfigurationIsUsed,
 				"ProgrammingLanguageConfiguration must not be null");
+		// SupervisedCode is the one record that carries the language, so it can and
+		// does validate its own names for that language rather than leaving it to the
+		// schema validator alone.
+		LanguageNameRules nameRules = theFollowingProgrammingLanguageConfigurationIsUsed.nameRules();
```
Claudia-Anthropica, line 72: @MarkusPaulsen [medium] The new language-aware constructor validates only the top-level package, main-class, and test-class names, while `PackagePermission` now checks only for null. A programmatic policy can therefore retain `new PackagePermission("com.foo.")`; `JavaArchunitTestCaseCollection` strips that trailing dot and grants `com.foo`, even though the file-policy path rejects the same selector. Validate every nested package permission through the selected `nameRules` here and add a programmatic-construction regression so this public API remains fail-closed.

In `src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/SupervisedCode.java`, resolving `nameRules` without applying it to `theFollowingResourceAccessesArePermitted.regardingPackageImports()` lets invalid programmatic package selectors survive and be normalised into real grants. Validate every nested package permission with the selected language rules before storing the resource accesses, and add a regression for a trailing-dot selector.

</details>

MarkusPaulsen:
@coderabbitai review

Three findings from the 2026-08-11 review are addressed in `dadd0549`. They were
posted as **outside diff range comments** in the review body rather than as inline
threads, so the sweep over unresolved `reviewThreads` did not see them. Noting that
here for the next reader, since a PR can look fully addressed while carrying them.

All three were still valid against the current file, which is otherwise untouched
by this branch:

- `docs/policy/SecurityPolicyManual.md` said `createRestrictive()` constructs a
  **10,000 ms** limit. Its own validation table on line 746 says 3,000, and so does
  the code. Corrected to 3,000 ms. This is the same stale figure Claudia found in
  `SecurityPolicyModel.puml`, so it existed in three places.
- **MD028**, a blank line between two block quotes.
- **MD031**, a fenced block with no blank line before it.

Both lint rules are fixed everywhere they occur in the file, not only at the lines
named: **three** MD028 sites and **twenty** MD031 sites. Fixing only the cited ones
would just return as the next finding.

One deviation from the suggested fix, deliberately. For MD028 the proposal was to
remove the blank line or make it quote-only; both merge two intentionally separate
call-outs into a single rendered box, which is a content change to a manual. An
empty HTML comment ends the first block quote for the linter **and** for the
renderer, so the two call-outs stay two call-outs. Every MD031 site is a top-level
paragraph followed by a fence, so the inserted blank line changes nothing that
renders.

CI is green on the previous head and the file carries no code.

src/main/java/de/tum/cit/ase/ares/api/policy/policySubComponents/SupervisedCode.java
MarkusPaulsen: Applied in `8f34f912`, with one correction to the reasoning and one consequence you did not name.

**The mechanism is exactly as described.** `SupervisedCode` resolved `nameRules` and
applied it to the supervised package, the main class and the test classes but not
to `regardingPackageImports`; `PackagePermission` only null-checks;
`PACKAGE_PATTERN` is `ident(\.ident)*`, so the file gate refuses `com.foo.`; and
[`JavaArchunitTestCaseCollection.java:233`](https://github.com/ls1intum/Ares2/blob/chore/improve-policy-subsystem/src/main/java/de/tum/cit/ase/ares/api/architecture/java/archunit/JavaArchunitTestCaseCollection.java#L233) strips the trailing dot. Verified each link
rather than taking the chain on trust.

**The impact is smaller than "turn an invalid selector into a real package
grant" suggests.** The grant that results is the one the author meant:
`com.foo.` normalises to `com.foo`, which is what the valid spelling grants. I
worked through the neighbouring shapes and none widens anything, `""`, `"."`,
`"com.foo.."` and `"com.foo.*"` all grant nothing or less. Nor is it reachable
from student code: the scanner-derived and essential packages already go through
[`JavaCreator.validatedAllowedPackage`](https://github.com/ls1intum/Ares2/blob/chore/improve-policy-subsystem/src/main/java/de/tum/cit/ase/ares/api/securitytest/java/creator/JavaCreator.java#L185), whose Javadoc states that placement
deliberately. The only unchecked path is a policy an instructor writes in Java.

So this is a consistency defect, not a sandbox weakening: the two ways of
expressing one policy disagreed, in a branch whose stated purpose is that they
agree. That is reason enough, and `ClassPermission` and `CommandPermission`
received the same treatment here.

**The consequence you did not name is the sharper one.** The same unvalidated
value also reaches [`JavaArchunitTestCase.java:64`](https://github.com/ls1intum/Ares2/blob/chore/improve-policy-subsystem/src/main/java/de/tum/cit/ase/ares/api/architecture/java/archunit/JavaArchunitTestCase.java#L64) and `JavaWalaTestCase:494`, which
interpolate it into **generated Java source** inside a string literal with no
escaping. It stays within instructor privilege, but it shows the gap is
`JavaCreator` line 150 passing the policy's own imports through unchecked, not
the record constructor.

**Fix.** The check is in `SupervisedCode`, as you proposed: it is the only place
holding both the language and the permissions. It needed a wildcard-tolerant
counterpart, so `LanguageNameRules` gains `requirePackageImport`, because `*` is a
legal package *import* and not a legal package *name*; `requirePackage` alone
would have refused it. Regressions in `PolicyValueContractTest` cover the
trailing dot, `com..foo`, `.com.foo`, `com.foo.*`, the empty string and a reserved
word, and that `*`, `java.util` and a bare identifier still pass.

458 policy tests and 484 including the creator and ArchUnit suites, all green,
and `mvn clean package` passes the SpotBugs and Javadoc gates. Recorded as
breaking change 9 in the description.

src/main/java/de/tum/cit/ase/ares/api/policy/PolicyValueValidator.java
```diff
@@ -0,0 +1,151 @@
+package de.tum.cit.ase.ares.api.policy;
```
Claudia-Anthropica, line 1: @MarkusPaulsen [medium] Moving this class deletes `de.tum.cit.ase.ares.api.policy.policySubComponents.PolicyValueValidator`, which is public in the local v2.1.0, v2.1.1, and v2.1.2 tags. Downstream imports and references to its released public patterns and methods therefore stop compiling; retain a deprecated compatibility facade in the old package for at least one cycle.

In `src/main/java/de/tum/cit/ase/ares/api/policy/PolicyValueValidator.java`, moving the public validator leaves no type at its released v2.1.x package and breaks downstream source compatibility. Add a deprecated forwarding class in `de.tum.cit.ase.ares.api.policy.policySubComponents` that preserves every released public field and method for at least one compatibility cycle.

</details>

MarkusPaulsen:
count towards it.

This description was written before those limits and exceeds 5 of them, so I have

<details><summary>Summary, as it read before (732 characters, limit 500)</summary>

Closes the places in the policy subsystem where an omission, an ambiguity or an unchecked value
was silently accepted, replaces the exact policy-version check with a supported range applied by
both gates, rewrites the failure messages so a policy author is told what to write, and moves the
two validators to packages that describe what they are. This branch also carries a self-contained
refactor of the Java security-policy director, which no longer inspects the concrete type of its
collaborators to decide whether to rebind them to the discovered build configuration. Each
collaborator now exposes a `withBuildConfiguration` method and decides for itself, and the policy
documentation gains three new files under `docs/policy/`.

</details>

<details><summary>1. Problem, as it read before (4025 characters, limit 1000)</summary>

Not one defect but one recurring shape, found by reading the policy value objects closely. In a
security boundary the direction a mistake fails in matters more than its size, and in several
places the direction was outwards.

**The policy version could not express what a release supports.** `SecurityPolicy` carried a
single `CURRENT_POLICY_VERSION` and demanded exact equality with it, in two independent places:
the record's canonical constructor and the schema validator. A release understands a *set* of
format versions whose two ends move independently, and nothing forced those sites to stay in
step. Introducing a version 2 would have invalidated every existing version 1 policy file, or, if
only one site were updated, made the reader and the record disagree about the same file.

**A network permission built without a port permitted every port.** The builder held the port in
an `int`, so an unset port was `0`, and `0` is not neutral here: it means every port.
`builder().onTheHost("example.com").sendData(true).build()` therefore granted sending across the
whole host. Every other field in the package defaults to the restrictive value, the booleans to
`false` and the permission lists to empty. This one defaulted to the most permissive value it
could hold.

**Resource accesses built without an execution limit ran unbounded.** `createRestrictive()`
deliberately adds a limit, with a comment explaining that an empty list would let a
non-terminating submission run forever. The plain builder had the same empty default and no such
compensation, so the two disagreed on what "nothing configured" means.

**Two policy values were never checked at all.** Every other value had to match a pattern, on both
the parse path and the construction path. `ClassPermission` and `CommandPermission` accepted
anything non-blank, in either path, so a malformed class name or a command carrying a line break
was taken at face value.

**A rejected value was reported by quoting the regular expression it failed.** For a host that
meant several hundred characters of IPv6 alternation, which states what the expression is without
saying what to write instead. It was also the only message family in the policy package that
bypassed localisation, so it stayed English while its neighbours were translated.

**A command permission printed as though it had no argument constraint.** `CommandPermission`
overrode `toString()` to return the command alone, a remnant of serialising it as a bare string
through `@JsonValue`. The annotation was removed when that was found to drop arguments from
written policies; the override outlived it. A permission for `git` with arguments `[status]`
still printed as `git`, in exactly the diagnostics someone reads to find out why a command was
refused.

**Two classes sat in packages that misdescribed them.** `PolicyValueValidator` lived among the
records in `policySubComponents` although it is not a value object, leaving the reader importing
patterns out of a package of permissions. `SecurityPolicySchemaValidator` lived in `reader.yaml`
although nothing in it is YAML: it validates a Jackson node tree, which is the same tree whichever
format produced it. That placement invited a duplicate for the next format.

Finally, the documentation was wrong in ways a reader would act on: two Javadoc blocks in
`SupervisedCode` had run together onto one line so their tags never rendered, one named a
parameter that does not exist, `ResourceLimitsPermission` still called itself `TimeoutPermission`,
`ThreadPermission` documented only fully qualified class names though the validator also accepts a
wildcard and six concurrency tokens, and nothing in the package documented what it throws.

The value-hardening root cause sits in the policy layer, in the value objects and their reader;
the separate director refactor described in §6 additionally touches `SecurityPolicyJavaDirector`
and its Java collaborators in `securitytest.java`. Neither part changes enforcement, AOP or
architecture behaviour.

</details>

<details><summary>2. Improvement from the user's perspective, as it read before (1143 characters, limit 1000)</summary>

- An instructor assembling a policy in code can no longer grant a whole host by forgetting a
port, or an unbounded run by forgetting a limit. Both now have to be written down.
- A rejected value now says what the field accepts and quotes the offending value:
`onTheHost must be a DNS name, an IPv4 or IPv6 address, localhost, or * for every host, but was
'not a host'`, in English and German, instead of the raw pattern in English only.
- A malformed class name or a command containing a control character is refused when the policy
is read, rather than accepted and then matching nothing during enforcement.
- A refused command is reported with its arguments, so a denial that hinges on the arguments can
be understood from the message rather than from the source.
- The Javadoc now describes the wildcard and the six concurrency tokens `ThreadPermission`
accepts, the placeholders a file path may use, and what every entry point throws.
- Once a version 2 of the policy format exists, an existing version 1 file keeps loading. That
compatibility is prepared here, not claimed here: the range is `[1, 1]`, so no policy file
changes behaviour.

</details>

<details><summary>3. Improvement from the maintainer's perspective, as it read before (2563 characters, limit 1000)</summary>

- One private helper is the single source of truth for the version invariant, so the constructor
and the builder cannot drift apart in either the condition or the message.
- The schema validator expresses the same invariants as the records, for the version and now for
commands too, so raising `MAXIMUM_POLICY_VERSION` is a one-line change rather than a hunt for
hard-coded equality.
- The remaining defaults across the package all fail closed, so "unset" has one meaning.
- The message wording lives beside the pattern it explains, and an unmapped pattern still
produces the old form rather than no message at all.
- `SecurityPolicySchemaValidator` now records why it is not an interface with one implementation
per format, so the next person does not build that seam where nothing varies.
- The package produces zero Javadoc warnings, down from one before this branch. The previous
`{@value}` tag pointed at `CURRENT_POLICY_VERSION`, which fails `maven-javadoc-plugin`,
configured here with `failOnError`; the tags now point at the range constants.
- The policy subsystem gains three new documentation files under `docs/policy/`:
`SecurityPolicyModel.puml` (the policy value-object model), `SecurityPolicyReaderAndDirectorClassDiagram.puml`
(the reader-and-director structure, including the new `ProgrammingLanguageConfiguration` enum and
its dependency on `SecurityPolicyDirector`), and `UMLClassDiagramRules.txt` (the conventions those
diagrams follow).

### Follow-up refactor: polymorphic collaborator configuration

`SecurityPolicyJavaDirector` chose whether to rebind a collaborator to the discovered
`BuildToolConfiguration` by inspecting its concrete type (`getClass() == JavaCreator.class`,
`instanceof JavaProgrammingExerciseProjectScanner`). That decision now lives in a
`withBuildConfiguration(BuildToolConfiguration)` method on `JavaCreator`, `JavaWriter` and
`JavaProjectScanner`, overridden in `JavaProgrammingExerciseProjectScanner`: an exact
framework-default type is rebound to the discovered configuration, and any subclass is returned
unchanged so a custom injected collaborator is never silently replaced. Scanner subclasses are now
preserved as well, consistent with the creator and writer, and this compatibility change is
documented on the method. Direct tests exercise exact-type rebinding, subclass preservation and
null rejection, and the reader-and-director class diagram (`docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml`)
adds the `ProgrammingLanguageConfiguration` enum and its dependency on `SecurityPolicyDirector`.

</details>

<details><summary>Breaking changes and migration, as it read before (7100 characters, limit 1000)</summary>

No policy file has to be rewritten and no generated security test code changes, so no instructor
has to touch an existing exercise. Nine changes affect code that constructs or inspects policy
objects programmatically, or the execution budget an exercise runs under, and they should be
weighed rather than waved through.

1. **`SecurityPolicy.CURRENT_POLICY_VERSION` is deprecated**, superseded by
`AresConstants.MINIMUM_POLICY_VERSION` and `AresConstants.MAXIMUM_POLICY_VERSION`, because one
constant cannot express a set of versions whose two ends move independently. An earlier revision
of this branch removed it outright; it shipped in **v2.1.0**, tagged 2026-07-21, so that broke a
released API and it is now retained as a `@Deprecated(forRemoval = true)` alias of the maximum,
which carries the value and the meaning it had. Migration is the rename, with one caveat the
Javadoc records: a comparison for *equality* against it has to become a range check, or it will
reject a version 1 policy once version 2 exists.
2. **`PolicyValueValidator` moves** from `…policy.policySubComponents` to `…policy`. It is a
public class that also shipped in v2.1.0, so any external import of it stops compiling.
Migration is the changed import. It was **not** moved to `api.util` on purpose: nine of its
constants and half its methods are policy vocabulary, encoding the implicit-thread-op tokens,
the recognised path placeholders and the eight configuration names, and all ten of its
consumers live inside `api.policy`. Moving it to `util` would put policy semantics in a
general utility package.
3. **`NetworkPermission.Builder.build()` now throws when the port was never set**, instead of
defaulting to `0` and thereby permitting every port. Callers relying on the default must pass
`0` explicitly. Every call site in this repository already sets the port.
4. **`ResourceAccesses.Builder` now starts with the restrictive execution limit** instead of an
empty list. A caller who wants no limit passes an empty list explicitly. This tightens the
result of any builder that omitted it.
5. **`CommandPermission.toString()` now returns the record's own representation**, including the
arguments, rather than the bare command. Nothing in this repository parses it; anything
downstream that did should read `executeTheCommand()` instead.
6. **`ClassPermission` and `CommandPermission` now reject values they previously accepted**: a
class name that is not a dot-separated Java name, and a command or argument containing a
control character, or a command with surrounding whitespace. Everything the repository's
fixtures and the twenty-two essential classes actually use was checked against the patterns
before adopting them.
7. **`JavaCreator`, `JavaWriter` and `JavaProjectScanner` gain a public `withBuildConfiguration`
method**, and custom *scanner* subclasses are now preserved rather than replaced by a
reconfigured plain scanner. A custom scanner subclass that needs the discovered configuration
must now override this method. No production caller is affected: the framework defaults behave
exactly as before.
8. **`ResourceLimitsPermission.createRestrictive()` returns 3,000 ms instead of 10,000 ms.** This
is the limit applied when a run has no policy at all, and together with item 4 it is now also the
default of `ResourceAccesses.Builder`, so it tightens the execution budget of every such run
rather than only of a builder that omitted the field. A submission that legitimately needs longer
than three seconds must declare a `regardingTimeouts` entry. Restoring the previous budget is one
line, `ResourceLimitsPermission.builder().withTimeout(10000).build()`, but the point of the
restrictive default is that a run with no policy should not be given ten seconds of unsupervised
time.
9. **`SupervisedCode` now rejects a package import that is neither a valid package name nor
`*`.** A policy read from a file was already held to this by the schema validator; a policy built
in code was held to it nowhere, because `PackagePermission` is language-agnostic and cannot know
what a package looks like, and `JavaCreator` validates only the imports it derives, not the ones
the policy supplies. `SupervisedCode` is the only place holding both the language and the
permissions, so the check goes there, beside the ones it already applies to the supervised
package, the main class and the test classes. `LanguageNameRules` gains
`requirePackageImport`, the wildcard-tolerant counterpart of `requirePackage`, since `*` is a
legal package *import* and not a legal package *name*. Every value the repository's fixtures and
the essential packages use was checked against the rule first. The values this newly refuses were
never usable as written: `com.foo.` reached enforcement as the grant `com.foo`, because the
ArchUnit matcher strips a trailing dot, and `com.foo.*` matched nothing at all.

Two further changes are deliberately **not** in that list, because both were reverted to stay
compatible after review.

`CommandPermission.fromJson` and `fromString` shipped in v2.1.0 and are **retained**, now
`@Deprecated(forRemoval = true)`, and with them the bare scalar command form `- git`. Removing
them was the earlier state of this branch, and it was wrong twice over: it dropped two public
methods from a released artefact, and it silently stopped loading version-1 policy files that use
the scalar form, while `MINIMUM_POLICY_VERSION` and `MAXIMUM_POLICY_VERSION` both still say 1. A
format cannot narrow without a version to narrow in. The form keeps the meaning it had, the
command with an **empty** argument list, and `SecurityPolicyStrictSchemaTest` pins that so it
cannot quietly widen to "any arguments"; the schema gate holds the scalar to `COMMAND_PATTERN`
exactly as it holds the mapping form. Both go with the next policy-format version.

`SecurityPolicyDirector.createTestCases(SecurityPolicy, Path, Path)` stays **concrete**. Declaring
it abstract, as this branch first did, breaks a director written against v2.1.0 at compile time
and, worse, with an `AbstractMethodError` for one already compiled. The reason it was made
abstract stands, though: the default it shipped with forwarded to the two-argument overload
unconditionally and so dropped `withinPath`, analysing the whole project when the caller asked for
one directory. The inherited default now forwards only the empty path, which is how an unscoped
analysis is written throughout the policy layer, and throws `UnsupportedOperationException` for
any real scope. A subclass that wants a scope overrides the method, as every director here does.

One further change is deliberately **not** in that list either. `SecurityPolicySchemaValidator` moves
from `…reader.yaml` to `…reader` and becomes public, but it was package-private before, so no
external code could reference it: the move is invisible outside the artefact and the new
visibility is an addition, not a break. It had to become public because its caller now sits in a
sub-package and can no longer see a package-private type; its Javadoc records that it is internal
to the reader.

</details>

docs/policy/SecurityPolicyModel.puml
MarkusPaulsen: Fixed in 7dd7ded0.

The note now reads `- createRestrictive(): 3,000 ms` rather than `- default: 3,000 ms`. Naming the factory rather than calling the value a default, because the same note says two lines further down that the timeout `must be set explicitly`, and the builder really does leave it at 0, which the compact constructor then rejects. Calling 3,000 a default would have swapped one inaccuracy for another.

docs/policy/SecurityPolicyReaderAndDirectorClassDiagram.puml
MarkusPaulsen: Fixed in 7dd7ded0. `{abstract}` is removed from the three-argument overload; the two-argument one above it keeps the marker, which matches `SecurityPolicyDirector:138` being abstract and `:175` being concrete for exactly the compatibility reason you name.

src/main/java/de/tum/cit/ase/ares/api/policy/PolicyValueValidator.java
MarkusPaulsen: Confirmed and fixed in 7dd7ded0. I checked the tags before acting: `policySubComponents/PolicyValueValidator.java` is present and `public final` in `v2.1.0`, `v2.1.1` and `v2.1.2`, publishing seven `public static final Pattern` constants and four public static methods.

`de.tum.cit.ase.ares.api.policy.policySubComponents.PolicyValueValidator` is restored as a `@Deprecated(forRemoval = true)` facade carrying that whole surface.

**It could not be a plain forward.** Only `FILE_PATH_PATTERN`, `HOST_PATTERN`, `matches` and `requireMatch` have a counterpart on the relocated validator. The five Java-specific members are now private regexes inside `JavaNameRules`, which is the point of this PR. Rather than recompile copies of those expressions in the facade, which would drift from the rules actually enforced and leave the facade advertising a contract Ares does not honour, `JavaNameRules` gained four package-private accessors. The facade is in the same package, so this needs no new public API, and the accessors are deleted together with the facade. `PolicyValueValidatorCompatibilityTest` asserts the constants are the *same objects*, so the no-drift property is tested rather than asserted in prose.

`PROGRAMMING_LANGUAGE_CONFIGURATION_PATTERN` is derived from `ProgrammingLanguageConfiguration.values()` instead of the released literal, so a configuration added to the enum cannot leave the constant behind.

**Three deliberate non-identities, none source breaking:**

1. `requireMatch` and `requirePackageImport` throw the released `IllegalArgumentException` but carry the localised wording. Reproducing the old English sentence would have the facade report a different reason than Ares itself.
2. `FILE_PATH_PATTERN` is the hardened constant that rejects control characters and Unicode line separators. Republishing the laxer released expression would let a caller pre-validate a path that enforcement then refuses.
3. The configuration pattern recognises the same eight strings; only its `pattern()` text differs.

The test pins the *released* surface, not the current implementation: every member is resolved through `getField`/`getMethod`, which see public members only, so demoting one fails the test instead of compiling quietly in the same package. The eight configuration names are written out literally rather than read from the enum that also builds the pattern, which would only have proved the constant agrees with itself.

Verified locally on Java 17: 485 policy tests pass, and checkstyle, PMD, SpotBugs and spotless are all clean.

MarkusPaulsen:
@coderabbitai review

docs/policy/SecurityPolicyManual.md
```diff
@@ -342,12 +345,14 @@ The `theSupervisedCodeUsesTheFollowingPackage` field specifies the root package
 **Examples:**
 
 Using a shallow package prefix:
+
 ```yaml
 theSupervisedCodeUsesTheFollowingPackage: "de.tum.cit.aet"
 ```
 This matches all classes under `de.tum.cit.aet` (including `de.tum.cit.aet.solution`, `de.tum.cit.aet.util`, etc.).
```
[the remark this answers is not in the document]
MarkusPaulsen: Fixed in 12401028, at every site rather than only the two named. A rule half-applied to one file just comes back as the next finding.

Sweeping the file turned up **four** MD031 sites, not one: lines 351, 358, 485 and 495, each a closing fence butted straight against the following paragraph or heading. And **three** MD040 sites, not one: lines 770, 802 and 808, all of them Ares exception output, now fenced as `text` as you suggested.

`SecurityPolicyManual.md` is the only Markdown file in this PR diff, and it now reports zero MD031 and zero MD040.

src/test/java/de/tum/cit/ase/ares/api/policy/policySubComponents/PolicyValueValidatorCompatibilityTest.java
```diff
@@ -0,0 +1,168 @@
...
+ * package up to 2.1.2.
+ * <p>
+ * Every member the facade republishes is exercised here, so removing or
+ * renaming one fails a test rather than silently breaking a downstream
+ * compilation that this repository never builds. The identity assertions
+ * additionally prove the facade forwards to the component that now owns each
+ * rule instead of holding a second copy of the expression, which is the failure
+ * mode a facade invites.
+ */
+@SuppressWarnings("deprecation")
```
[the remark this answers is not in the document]
MarkusPaulsen: Correct, fixed in 12401028. The annotation is now `@SuppressWarnings({ "deprecation", "removal" })`.

You are right that the two lints are separate: the facade is `@Deprecated(forRemoval = true)`, and javac reports uses of a forRemoval API under `removal`, which `deprecation` does not cover. Since this test exists precisely to exercise the deprecated surface, it touches it on nearly every line, so the suppression had to cover both. I left a comment on the annotation recording why, so it does not get trimmed back to one key later.

MarkusPaulsen:
### Outside-diff findings from the 2026-08-14 review

These three carry no review thread, so recording here that all are fixed in [`12401028`](https://github.com/ls1intum/Ares2/commit/12401028).

**`JavaNameRules.java:35-40` — identifier-ignorable characters (Major).** Confirmed and fixed. I probed it rather than taking it on trust: `\p{javaJavaIdentifierPart}` accepted **all ten** ignorable code points I tested, and `"de.tum" + U+0000 + "cit"` passed as a valid package name. Also U+007F, the C1 controls, and the format characters U+00AD, U+200B, U+200D and U+FEFF.

This matters beyond identifier equality. A name is not merely compared, it is written into generated sources, into AspectJ and ArchUnit configuration and into failure messages. A NUL corrupts an artefact that assumes one value per line, and a zero-width character produces two allowlist entries a reader cannot distinguish. `PolicyValueValidator` already excludes exactly these characters from paths and commands, so permitting them in names had applied the rule to two thirds of the policy.

Only the identifier *part* is constrained, since `Character.isJavaIdentifierStart` accepts no ignorable character to begin with. I verified the exclusion keeps legitimate identifiers, including non-ASCII ones such as `de.übung.例` and `straße`.

The new test asserts `Character.isJavaIdentifierPart(codePoint)` holds *before* asserting the rules reject it, so it cannot pass vacuously. Coverage is package, package import, type name, class path and thread construct. I also scanned every resource and fixture in the repository: none carries such a character, so no existing input changes meaning. The narrowing reaches the four republished name constants on the compatibility facade, documented there next to the equivalent note on `FILE_PATH_PATTERN`.

**`SupervisedCode.java:302-306`.** Fixed, now "the shape the selected programming language requires of it". Verified that `build()` does validate through the selected `ProgrammingLanguageConfiguration`, which is the point of the language split.

**`PolicyValueContractTest.java:29-30`.** Fixed. Verified the claim: `SecurityPolicy.Builder.thisPolicyFileCompliesToThePolicyVersion` delegates to `requireSupportedPolicyVersion` at `SecurityPolicy.java:194`, and the javadoc at `:72` says the two entry points "cannot drift apart". The comment claimed the opposite, so it now describes the assertions as covering the second entry point to the shared check.

495 policy tests pass locally on Java 17; checkstyle, PMD, SpotBugs and spotless are clean.

MarkusPaulsen:
@coderabbitai review

MarkusPaulsen:
### On the `program` → `programme` suggestion

I have not made that change, because it would be incorrect. British English keeps the spelling `program` for this one sense: software is a **program**, whereas a **programme** is a broadcast, an event schedule or a course of study. Oxford, Cambridge and the Guardian style guide all carry that carve-out. "A programme name or a path to one" is not something a British English style guide would ask for.

Fixed differently in [`bfe797c8`](https://github.com/ls1intum/Ares2/commit/bfe797c8): `program` was not the precise word there either. The pattern matches the name of an executable or a path to one, so both places now say **executable**. That is unambiguous, correct in either dialect, and will not attract the same finding again.

I changed both occurrences, the `COMMAND_PATTERN` javadoc in `PolicyValueValidator` and the mirrored comment in `PolicyValueContractTest`, since fixing only the flagged one would leave the two explanations of the same pattern disagreeing.

One note for the repository rather than for this PR: `.coderabbit.yaml` lists `program` among the American spellings to flag, alongside `behavior`, `color` and `initialize`. Those four are not alike. The instruction will keep producing this false positive on any prose about running an executable, so it is worth carving out the computing sense there.

@coderabbitai review

Claudia-Anthropica (dismissed):
@MarkusPaulsen The exact head preserves the released compatibility surface and adds focused validation and regression coverage for the changed policy contracts. [low] The testing manual still says `new ClassPermission("not a class name")` throws, although the constructor accepts non-blank names and Java-specific validation occurs in `JavaCreator`; this does not block approval. The snapshot has no required checks configured and no relevant completed failure.

MarkusPaulsen:
@coderabbitai review

MarkusPaulsen:
@coderabbitai full review

Claudia-Anthropica (changes requested):
@MarkusPaulsen [medium] This head removes the public `SecurityPolicyReaderAndDirector.factoryAndBuilder()` accessor, breaking source compatibility for downstream users. The method exists in each local v2.1.0–v2.1.2 release tag, and the updated integration test now uses reflection to access the same field. Restore the accessor, marking it deprecated if removal is intended after a compatibility cycle.

Claudia-Anthropica (changes requested):
@MarkusPaulsen [medium] This head still removes the public `SecurityPolicyReaderAndDirector.factoryAndBuilder()` method: it exists in every local v2.1.0–v2.1.2 tag and in the captured merge base, while the current class provides no replacement. `SecurityPolicyPipelineIntegrationTest` now accesses the same private field reflectively, confirming that the behaviour remains but its released accessor was deleted; restore the accessor for at least one compatibility cycle, marking it deprecated if removal is intended. Several non-required CI jobs were still running in the snapshot, with no failed check captured.

Claudia-Anthropica (changes requested):
@MarkusPaulsen [medium] The released `SecurityPolicyReaderAndDirector.factoryAndBuilder()` accessor is still absent on this exact head: it exists in the merge base and v2.1.0–v2.1.2, while `SecurityPolicyPipelineIntegrationTest` still uses reflection to inspect the same private factory. Restore the public accessor for a compatibility cycle, deprecated if desired; `documentation / Build Docusaurus site` is also red, but the captured evidence contains only exit code 1 and no failing step or report, so I could not attribute it to this change.

Claudia-Anthropica (changes requested):
@MarkusPaulsen [medium] The released `SecurityPolicyReaderAndDirector.factoryAndBuilder()` accessor remains absent on this exact head even though it exists in v2.1.0–v2.1.2 and the merge base; `SecurityPolicyPipelineIntegrationTest` still reflects into `securityTestCaseFactoryAndBuilder`, so downstream callers lose source compatibility. Restore the public accessor for a compatibility cycle, deprecated if desired; `documentation / Build Docusaurus site` is also red, but the snapshot contains only exit code 1 and no failing step or report, so I could not attribute it.

MarkusPaulsen:
Zu Claudias offenem Befund: `SecurityPolicyReaderAndDirector.factoryAndBuilder()` ist in 1ffcee49 wieder da, jetzt als `@Deprecated(since = "2.1.5", forRemoval = true)`, damit ein Aufruf aus einem Übungs-Repository weiterhin kompiliert. `SecurityPolicyPipelineIntegrationTest` benutzt den Accessor wieder statt Reflection auf das private Feld.

Dazu zwei Dinge aus dem Merge von main: Der Renovate/Dependabot-Absatz in AGENTS.md stand doppelt, weil beide Seiten ihn ergänzt hatten (f786505a), und vier Doku-Seiten von main nannten noch das alte restriktive Limit von 10.000 ms statt der 3.000 ms dieses PRs (44096d26, 710b5219).

Der Diagramm-Check der Doku-CI ist jetzt reproduzierbar: PlantUML misst Textbreiten mit den Schriften des ausführenden Rechners, deshalb kann ein SVG von macOS nie zu dem des Ubuntu-Runners passen. Der Check rendert nach `documentation/.plantuml-check`, behält den Baum bei einer Abweichung, und die CI lädt ihn als Artefakt `plantuml-renders` hoch (c760c042, c124c566, cb7bf9a9). Die committeten SVGs stammen genau aus diesem Artefakt.

Claudia-Anthropica (approved):
@MarkusPaulsen The exact head restores the released `factoryAndBuilder()` accessor and the integration test now exercises it directly, resolving the prior compatibility finding. The current snapshot contains no failed checks; two non-required WALA integration jobs remain in progress. I found no unresolved medium-or-higher defect.

PR #152, opened 2026-07-23
fix(phobos): correct copy helper quoting

ShudongCai:
## Summary

Correct the `PhobosCopyTool.sh` template so its target-directory assignment uses ASCII double quotes instead of Unicode curly quotation marks, and pin the fix with a regression test. Without this, the copy step in every generated Phobos exercise fails. (Audit finding I-002.)

## Linked issues

None. I-002 comes from an internal audit report, not a GitHub issue.

## 1. Problem

Observed under any configuration that generates a Phobos exercise (the script is a static template resource, so no Java version, build tool, AOP mode or architecture mode changes the defect). `PhobosCopyTool.sh` assigned the target directory with Unicode curly quotation marks:

```bash
TARGET_DIR=“/var/tmp/opt/core”
```

Bash does not treat `“` (U+201C) and `”` (U+201D) as quoting operators, so they stay literal in the value. `TARGET_DIR` resolves to `“/var/tmp/opt/core”` rather than `/var/tmp/opt/core`, and the following `cp -v SpecificExercise.cfg "$TARGET_DIR/"` then copies into a directory that does not exist and fails (the script also runs under `set -e`, so the failure aborts it).

The root cause sits in the build-integration layer: this script is copied verbatim into generated exercise artifacts (`PhobosCopyFiles.csv` -> `JavaWriter`), so the defect propagates into every generated exercise rather than staying local to the template. It is not a false negative or false positive of the security enforcement itself. It is a setup step that fails closed in the wrong place, breaking a correct exercise.

An instructor generating a Phobos exercise gets a copy step that works: `SpecificExercise.cfg` is copied into `/var/tmp/opt/core/` as intended, instead of the exercise breaking at the `cp` into a non-existent `“/var/tmp/opt/core”` directory.

The authoritative template is now pinned by `PhobosShellContractTest.copyToolAssignsTargetDirectoryWithAsciiQuotes`, which parses the assignment in Java and rejects any reappearance of the curly quotation marks, so this class of typo cannot silently return. The check is deliberately self-contained: it never executes the fixture and never interpolates its path into a shell command (it reads the file, matches the assignment with a regex, and validates syntax by feeding the content to `bash -n` on stdin).

## 4. Testing manual

**Prerequisites**

1. JDK and Maven as for any Ares 2 build, plus a `bash` on `PATH` (the test shells out to `bash -n`). No exercise repository, policy file or echo server is needed: the fixture is the packaged template resource `src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/PhobosCopyTool.sh`.
2. Check out this branch.

**Steps**

1. `mvn -o test -Dtest=PhobosShellContractTest -f pom.xml`
2. Optionally, open `PhobosCopyTool.sh` and confirm the assignment reads `TARGET_DIR="/var/tmp/opt/core"`.

**Expected result**

- `copyToolAssignsTargetDirectoryWithAsciiQuotes` passes: the script contains no `“`/`”`, contains exactly one `TARGET_DIR=` assignment, that assignment matches `^TARGET_DIR="(/var/tmp/opt/core)"$`, and `bash -n` on the script content exits 0.

- Reintroducing the curly quotation marks (or changing the value, or adding a second `TARGET_DIR=` line) makes the test fail: the `assertFalse` on U+201C/U+201D, the exactly-one-assignment `assertEquals(1, ...)`, and the anchored regex each reject the broken form. The test was confirmed to fail against the original curly-quote script before the fix.

**Modes exercised**

- [ ] ArchUnit + AspectJ
- [ ] WALA + AspectJ
- [ ] WALA + instrumentation

No production Java changed. The corrected artefact is the shell-script template resource `PhobosCopyTool.sh`; the only Java added is the regression test `PhobosShellContractTest`, which pins that resource. There is therefore no production class with Instruction/Branch/Line/Method coverage to report.

None. The public API under `de.tum.cit.ase.ares.api`, the security policy file format, the generated security test code and the minimum JDK/Maven/Gradle versions are all unchanged. The only behavioural change is that the copy step in a newly generated exercise now succeeds where it previously failed.

## Checklist

## Review progress

- [ ] Code review
- [ ] Manual test

### DIFF
--- a/src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/PhobosCopyTool.sh
+++ b/src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/PhobosCopyTool.sh
@@ -2,6 +2,6 @@
 # Script to copy required files to /var/tmp/opt/core/
 set -e  # Exit immediately if a command exits with a non-zero status
 set -u  # Treat unset variables as an error
-TARGET_DIR=“/var/tmp/opt/core”
+TARGET_DIR="/var/tmp/opt/core"
 # Copy SpecificExercise.cfg to target directory
 cp -v SpecificExercise.cfg "$TARGET_DIR/"
--- a/src/test/java/de/tum/cit/ase/ares/api/phobos/PhobosShellContractTest.java
+++ b/src/test/java/de/tum/cit/ase/ares/api/phobos/PhobosShellContractTest.java
@@ -1,6 +1,7 @@
 package de.tum.cit.ase.ares.api.phobos;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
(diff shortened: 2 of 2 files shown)


MarkusPaulsen:
@coderabbitai review

Claudia-Anthropica (dismissed):
@ShudongCai The template now uses ASCII double quotes for the intended target path, and the regression test safely validates the exact assignment and Bash syntax without executing fixture content. The prior shell-boundary concern is resolved, and all captured CI checks succeeded on this head.

MarkusPaulsen:
count towards it.

This description was written before those limits and exceeds 1 of them, so I have

<details><summary>1. Problem, as it read before (1119 characters, limit 1000)</summary>

Observed under any configuration that generates a Phobos exercise (the script is a static template resource, so no Java version, build tool, AOP mode or architecture mode changes the defect). `PhobosCopyTool.sh` assigned the target directory with Unicode curly quotation marks:

```bash
TARGET_DIR=“/var/tmp/opt/core”
```

Bash does not treat `“` (U+201C) and `”` (U+201D) as quoting operators, so they stay literal in the value. `TARGET_DIR` resolves to `“/var/tmp/opt/core”` rather than `/var/tmp/opt/core`, and the following `cp -v SpecificExercise.cfg "$TARGET_DIR/"` then copies into a directory that does not exist and fails (the script also runs under `set -e`, so the failure aborts it).

The root cause sits in the build-integration layer: this script is copied verbatim into generated exercise artifacts (`PhobosCopyFiles.csv` -> `JavaWriter`), so the defect propagates into every generated exercise rather than staying local to the template. It is not a false negative or false positive of the security enforcement itself. It is a setup step that fails closed in the wrong place, breaking a correct exercise.

</details>

Claudia-Anthropica (approved):
@ShudongCai The ASCII quote correction makes `TARGET_DIR` resolve to the intended directory. The regression test safely verifies the assignment and Bash syntax without executing fixture content, and all captured checks succeeded.

PR #164, opened 2026-08-02
Report which classes the no-policy scan recognised as tests

MarkusPaulsen:
## Summary

Report how many classes the no-policy scan recognised as test classes, so that the derived exempt set can be observed at all.

## Linked issues

Follows up on review feedback on #158, where a reviewer was asked to "note which classes Ares reports as test classes" and could not find anywhere that Ares reports them.

## 1. Problem

Only the no-policy path derives the exempt test classes from project code. `TestCaseAbstractFactoryAndBuilder` reads `theFollowingClassesAreTestClasses` when a policy is present, and calls `projectScanner.scanForTestClasses()` only when one is not.

That derived set was completely invisible. `JavaProjectScanner` had no logger at all, so an exercise author whose test was unexpectedly supervised, or unexpectedly exempt, had nothing to inspect but the JUnit XML report, which shows what *ran* rather than what Ares *trusted*. Those are different questions, and the second one is the one that matters when the no-policy fallback misbehaves.

The root cause sits in the project-scanning layer, upstream of enforcement. It is a diagnostics gap rather than a defect: no forbidden code is let through and no correct submission is failed, but the fallback cannot be audited.

An instructor using the no-policy fallback can now see that the scan ran and how many classes it recognised, and can raise the level for this one logger to see which. Previously the only way to reason about the exempt set was to infer it from which tests executed.

The no-policy path becomes auditable in CI output, so a future report of "my test was skipped" or "this class was exempt" can be answered from a log line rather than by reconstructing the scan by hand.

## Why the count is at INFO and the names are not

The count is harmless wherever build output ends up. The fully-qualified class names are not: a build log that reaches students would disclose the hidden-test structure. The names therefore go to TRACE, which the shipped `src/main/resources/logback.xml` leaves off, its root level being `debug`. DEBUG would not have been sufficient for the same reason. An instructor who needs the names raises the level for `de.tum.cit.ase.ares.api.securitytest.java.projectScanner.JavaProjectScanner` alone.

## 4. Testing manual

Scenario: *I have an Ares 2 exercise* that runs locally, with **no** security policy file, so the no-policy path is taken. No Artemis instance is involved. The change is build-tool independent; the commands below use Maven, and Gradle behaves identically.

**Prerequisites**

1. JDK 17 and Maven. No Artemis, database or network service.
2. Ares built and installed from this branch: `mvn -o install -DskipTests`.
3. A minimal exercise project depending on the Ares built above, with no policy file. `examples/ares-exercise-maven` on the `docs/setup-manual-and-exercise-template` branch is such a project once its `@Policy` annotation is removed; otherwise any throwaway project with one supervised class and one JUnit test class will do.

**Steps**

1. Run `mvn test` in the exercise project.
   *Expected:* the output contains exactly one line `No-policy scan recognised N test class(es).` at `[INFO]`, where N is the number of test classes in the project. The class names do **not** appear.
2. Re-run with the names enabled, for example by adding a logger entry at level `TRACE` for `de.tum.cit.ase.ares.api.securitytest.java.projectScanner.JavaProjectScanner` in the exercise's own logback configuration.
   *Expected:* the INFO count is still emitted, and a second line at `[TRACE]` lists the fully-qualified names, sorted. TRACE adds to the count, it does not replace it.
3. Add a `SecurityPolicy.yaml` and a `@Policy` annotation to the test, then run `mvn test` again.
   *Expected:* neither line appears, because the policy path never consults the scanner.

**Expected result**

The no-policy fallback states how many classes it recognised, and states which ones only when explicitly asked. Before this change neither was available at any level.

This change adds a diagnostic and alters no decision, so nothing that was previously blocked becomes permitted. Two properties must nonetheless hold, and both are covered by tests:

- The names must not be emitted at the shipped level. `JavaProjectScannerDiagnosticsTest.reportsCountAndWithholdsNames` pins the logger to DEBUG, the level `logback.xml` gives it, and asserts that no event contains a class name.
- The policy path must not scan at all. `TestCaseAbstractFactoryAndBuilderTest.exemptClassesComeFromPolicyNotScan` already asserts `verify(mockProjectScanner, never()).scanForTestClasses()`, which is a stronger guarantee than asserting the absence of a log line, so no new test was added for it.

**Modes exercised**

- [ ] ArchUnit + AspectJ
- [ ] WALA + AspectJ
- [ ] WALA + instrumentation

| Class | Line coverage | Confirmation (meaningful assertions) |
| --- | ---: | :---: |
| `JavaProjectScanner` | the added `reportRecognisedTestClasses` is fully covered by the two new tests | yes — `JavaProjectScannerDiagnosticsTest` captures the logger with a Logback `ListAppender` and asserts the event count, the level of each event, that the count appears in the message, and that the names appear at TRACE and only at TRACE |

None to the public API under `de.tum.cit.ase.ares.api`, the policy file format, the generated security test code, or the minimum JDK, Maven or Gradle version.

**Release note:** this must ship in a release **after** 2.1.1. `v2.1.1` is already tagged at `b26fd151` and Maven Central releases are immutable, so this cannot be published under that version. The setup manual deliberately does not document the new log line yet; that paragraph is added once the release carrying it is on Central.

## Checklist

- [ ] Documentation (`docs/`, `README.adoc`, Javadoc) was updated where the change is user-facing.

## Review progress

- [ ] Code review
- [ ] Manual test

### DIFF
--- a/src/main/java/de/tum/cit/ase/ares/api/securitytest/java/projectScanner/JavaProjectScanner.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/securitytest/java/projectScanner/JavaProjectScanner.java
@@ -17,6 +17,9 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nonnull;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.github.javaparser.JavaParser;
 import com.github.javaparser.ParseProblemException;
 import com.github.javaparser.ParserConfiguration;
@@ -35,6 +38,7 @@ import de.tum.cit.ase.ares.api.util.ProjectSourcesFinder;
 
 /** JavaParser-backed, deterministic Java project scanner. */
 public class JavaProjectScanner implements ProjectScanner {
+	private static final Logger LOG = LoggerFactory.getLogger(JavaProjectScanner.class);
 	// The test annotations Ares recognises, as fully-qualified type names. An
 	// annotation use counts only when it resolves to one of these types: a
 	// fully-qualified use names the type outright, and a bare simple name is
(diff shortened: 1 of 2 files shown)


MarkusPaulsen:
@coderabbitai full review

PR #167, opened 2026-08-03
Make the WALA backend work without jqwik and let AspectJ report the resolved runtime declaration

MarkusPaulsen:
## Summary

Makes the WALA architecture mode usable in any consumer that does not itself depend on jqwik, where it previously failed to initialise at all, and lets AspectJ denial messages name the implementation that instrumentation names, so one expectation file can serve both AOP modes.

## Linked issues

None. Both defects were found by the CI matrix of `ls1intum/SCOREReproducibilityPackage`, which runs Ares across all four mode combinations; no issue was filed before this pull request.

## 1. Problem

Two unrelated defects, both present in released 2.1.0 and 2.1.1, both surfaced by the same run. Of the twelve matrix jobs there (four modes x three JDKs), only ArchUnit + instrumentation was green.

**a. WALA is dead on arrival in consumers without jqwik (false positive, enforcement layer).**

Observed on JDK 17, 21 and 25, Gradle, both AOP modes, whenever the architecture mode is WALA:

```
java.lang.NoClassDefFoundError: net/jqwik/api/Property
  at ...wala.CustomCallgraphBuilder.<clinit>(CustomCallgraphBuilder.java:86)
  at ...wala.JavaWalaTestCase.classpathFingerprint(JavaWalaTestCase.java:342)
Caused by: java.lang.ClassNotFoundException: net.jqwik.api.Property
```

Expected instead: the analysis runs and the policy is enforced.

Root cause: `CustomCallgraphBuilder` built its trusted-framework origin set from 21 class literals in a static initialiser. jqwik is the only `provided`-scope entry among them, so it is absent from any consumer that does not depend on it directly. The literals are resolved while the varargs array is built, which is *before* `trustedFrameworkCodeSources` is entered, so the defensive catch inside that method is unreachable for this failure: `NoClassDefFoundError` is an `Error`, not a `RuntimeException`, and it is thrown at the call site. The class is then permanently erroneous, so one absent JAR produced 221 identical failures in a single job.

This is a **false positive**: correct submissions cannot be graded at all under WALA, because Ares aborts before analysing anything. It sits in the enforcement (architecture) layer.

**b. The two AOP modes name the same blocked call differently (usability defect, enforcement layer).**

Observed under ArchUnit + AspectJ on all three JDKs. The block always worked and the target was always correct; only the reported class differed:

| Instrumentation reports | AspectJ reports |
| --- | --- |
| `javax.net.DefaultSocketFactory.createSocket` | `javax.net.SocketFactory.createSocket` |
| `sun.security.ssl.SSLSocketFactoryImpl.createSocket` | `javax.net.ssl.SSLSocketFactory.createSocket` |
| `sun.nio.ch.AsynchronousSocketChannelImpl.connect` | `java.nio.channels.AsynchronousSocketChannel.connect` |
| `sun.nio.ch.DatagramChannelImpl.connect` | `java.nio.channels.DatagramChannel.connect` |
| `Executors$DelegatedScheduledExecutorService.schedule` | `java.util.concurrent.ScheduledExecutorService.schedule` |
| `java.util.stream.AbstractPipeline.parallel` | `java.util.stream.Stream.parallel` |

Root cause: AspectJ weaves `call(...)` join points in the supervised code, so `thisJoinPoint.getSignature()` names the **static receiver type at the call site**. Instrumentation advises the implementation whose bytecode runs and reports it through `@Advice.Origin("#t")`. AspectJ never observes `sun.nio.ch.DatagramChannelImpl` at all. Both are correct; they name the same operation at different levels of the hierarchy.

This is **neither a false positive nor a false negative**: enforcement is identical in both modes. It is a diagnostics gap, but a consequential one, because an expectation authored against one mode can never match the other, which makes a cross-mode comparison impossible to express.

- Instructors can actually ship an exercise on the WALA architecture mode. Previously any exercise repository without a direct jqwik dependency failed before the first rule ran, and the message named a library the instructor never referenced.
- A denial message now carries both identities: the API the student's code called, and the implementation that actually ran. Previously an instructor reading an AspectJ failure saw only the interface, which is often not the class documented in a stack trace elsewhere.
- One expectation file can serve both AOP modes, so an exercise no longer needs mode-specific expected output.

- Optionality is now explicit and contained. Required compile-scope frameworks keep their class literals and their compile-time checking; only genuinely optional ones are resolved by name, so the next `provided`-scope dependency added to that list cannot silently reintroduce this defect.
- Trust is resolved per analysis rather than in a global static, and `dependencyFingerprint` is now explicitly excluded from optional trust, which removes a path where an unvalidated origin could have dropped student bytecode from the verdict cache key.
- `buildCallGraph` now rejects a classpath the builder was not constructed for on the *first* call, not only once a graph has been cached. Every current caller passes the same value to both, so this was unreachable, but the scope, hierarchy and trust boundary are all fixed at construction and the guard did not say so.
- Two mode-comparison behaviours are now pinned by tests instead of being implicit.

## 4. Testing manual

Scenario: *I have an Ares 2 exercise.* No Artemis instance is involved.

**Prerequisites**

1. JDK 17 (Temurin) and Maven 3.8+.
2. This branch built and installed: `mvn -o install -DskipTests`.
3. An exercise whose policy uses `JAVA_USING_MAVEN_WALA_AND_INSTRUMENTATION` (for part a) and one using `JAVA_USING_MAVEN_ARCHUNIT_AND_ASPECTJ` (for part b). The forbidden-access subjects in `src/test/java/de/tum/cit/ase/ares/integration/aop/forbidden/subject/` serve as ready-made examples.
4. For network tests, the echo server described in `AGENTS.md`.

**Steps**

1. *Part a, the WALA regression.* In an exercise consuming Ares from step 2, ensure jqwik is **not** on the test classpath (it is `provided`-scope, so this is the default unless declared). Run the exercise's tests under a WALA policy.
2. Reproduce the old behaviour by checking out `main` in place of this branch, installing it, and repeating step 1.
3. *Part a, isolated.* `mvn -o test -Dtest=CustomCallgraphBuilderTest` in this repository. `testCallgraphBuilderInitialisesWithoutOptionalJqwikOnTheClasspath` builds a `URLClassLoader` over the real test classpath with the jqwik entries removed and loads `CustomCallgraphBuilder` through it.
4. *Part b.* `mvn -o test -Pfunctional-tests -Dtest=AspectJDeclarationNormalisationTest` in this repository. It drives one forbidden `SocketChannel.connect` through a real woven call under both an AspectJ and an instrumentation policy.

**Expected result**

- Step 1: the analysis runs and the policy is enforced; no `NoClassDefFoundError`.
- Step 2: the run fails immediately with `NoClassDefFoundError: net/jqwik/api/Property` from `CustomCallgraphBuilder.<clinit>`, repeated for every test.
- Step 3: 29 tests pass.
- Step 4: 2 tests pass. The messages the assertions pin are:

```
AspectJ:         via java.nio.channels.SocketChannel.connect(java.net.SocketAddress) [resolved runtime declaration: sun.nio.ch.SocketChannelImpl.connect(java.net.SocketAddress)]
Instrumentation: via sun.nio.ch.SocketChannelImpl.connect(java.net.SocketAddress)
```

This change must not make Ares more permissive anywhere, and enforcement is deliberately untouched.

1. Every forbidden-access integration test must still fail with a `SecurityException`: `mvn -o test -Pfunctional-tests -Dtest='de.tum.cit.ase.ares.integration.aop.forbidden.*Test'`. A permissive regression shows up as a test that no longer throws.
2. Enforcement keys must still be the **static** signature, not the resolved one. The ignore maps are keyed on `java.lang.Runtime.exec`, `java.lang.ProcessBuilder.start` and `java.io.File.delete`, and `java.io.File` is not final, so had the resolved name been substituted a student subclass overriding `delete()` could have shifted the key, missed the map and changed which fields are inspected. Confirm with `grep -n "extractMethodNameWithoutModifiers(fullMethodSignature)"` across the four advice files: all seven occurrences must use the raw `fullMethodSignature`, and all fourteen `describeDeniedCall` call sites must sit inside a `throw new SecurityException(localize(...))`.
3. A forged optional-framework origin must never be trusted: `testOptionalFrameworkOriginInsideSupervisedProjectIsRejected` places a candidate inside the supervised project's own output root and asserts it is not admitted.
4. A partially present optional framework must contribute nothing: `testOptionalFrameworkOriginsAreDiscardedWhenAnyClassIsAbsent`.
5. Application-loaded receivers must never be reflected over inside the advice guard: `describeDeniedCall_fallsBackForApplicationLoadedTarget`, confirmed end to end by a real run where a `TestHttpClient` receiver produces no appended clause.

**Modes exercised**

- [x] ArchUnit + AspectJ
- [x] ArchUnit + instrumentation
- [x] WALA + AspectJ
- [x] WALA + instrumentation

All four were exercised: the forbidden-access suite alone carries 172, 172, 171 and 171 policies for the four combinations respectively, and the whole suite ran green (1105 tests, 0 failures) on JDK 17. The new `AspectJDeclarationNormalisationTest` itself covers only ArchUnit + AspectJ and ArchUnit + instrumentation, because the behaviour it pins is a property of the AOP mode and is independent of the architecture mode.

Read from `site/jacoco/jacoco.csv` in the `coverage-report` artefact of the Coverage Report job on run [30809788300](https://github.com/ls1intum/Ares2/actions/runs/30809788300) (commit `eb23cfd0`), which merges the JaCoCo execution data of every test job. All five counters are reported per #170; the table previously carried line coverage only.

| Class | Instruction coverage | Branch coverage | Line coverage | Complexity coverage | Method coverage | Confirmation (meaningful assertions) |
| --- | ---: | ---: | ---: | ---: | ---: | :---: |
| `CustomCallgraphBuilder` | 86.2% (1550/1798) | 74.7% (133/178) | 81.0% (272/336) | 68.5% (98/143) | 94.4% (51/54) | yes |
| `JavaAspectJThreadSystemAdviceDefinitions` | 70.6% (878/1244) | 67.5% (164/243) | 73.1% (201/275) | 55.3% (84/152) | 96.7% (29/30) | yes |
| `JavaAspectJFileSystemAdviceDefinitions` | 72.4% (1644/2272) | 56.4% (261/463) | 71.8% (367/511) | 44.9% (123/274) | 92.3% (36/39) | yes |
| `JavaAspectJCommandSystemAdviceDefinitions` | 66.2% (1038/1569) | 53.1% (187/352) | 68.2% (225/330) | 37.5% (78/208) | 93.8% (30/32) | yes |
| `JavaAspectJAbstractAdviceDefinitions` | 63.0% (700/1111) | 66.2% (104/157) | 65.1% (181/278) | 61.8% (68/110) | 90.0% (27/30) | yes |
| `JavaAspectJNetworkSystemAdviceDefinitions` | 53.5% (772/1442) | 50.2% (146/291) | 56.1% (180/321) | 34.9% (61/175) | 70.4% (19/27) | yes |

The four `*SystemAdviceDefinitions` classes change only at their denial throw sites. Those lines are covered by the forbidden-access suite, which asserts on the thrown `SecurityException` and its message rather than merely executing the path, and the two new tests assert the message text directly.

`CustomCallgraphBuilder` carries the larger share of new logic; its optional-origin resolution, all-or-nothing composition, supervised-root rejection and the isolated-classloader initialisation path each have a dedicated test.

The wider table makes a pre-existing gap visible, and it is worth stating plainly rather than leaving it to be read out of the numbers. Branch and complexity coverage sit well below line coverage across the four advice classes, lowest at `JavaAspectJNetworkSystemAdviceDefinitions` with 50.2% of branches and 34.9% of complexity. This is not introduced here: the advice files are generated-shaped pointcut and guard code whose untaken branches are the per-API dispatch arms that no forbidden-access subject exercises, and this pull request changes only the throw sites, which are covered. It does mean the denial paths of this repository are less exhaustively pinned than the line figures alone suggested, and closing that belongs in its own pull request against the forbidden-access subject set rather than here.

No change to the public API under `de.tum.cit.ase.ares.api`, the policy file format, the generated security test code, or the minimum JDK, Maven or Gradle version.

One behaviour change is visible to consumers: **AspectJ denial messages may now carry a trailing `[resolved runtime declaration: ...]` clause.** The change is additive and the guaranteed static signature stays first, so any assertion that matches a substring of the message continues to match. An exercise that compares the *entire* message for exact equality against a recorded string will need that string updated. No migration is needed otherwise.

The version is set to 2.1.2 and is **not** released. Consumers pinned to 2.1.0 or 2.1.1 need that release to pick either fix up; until then, adding jqwik as a test dependency is a workaround for the WALA defect only.

## Known limitations, deliberately not addressed here

- `RESERVED_PACKAGE_PREFIXES` does not cover several namespaces that Ares trusts or excludes **by name** (`net/jqwik`, `org/junit`, `org/mockito`, `org/hamcrest`, `org/assertj`, `org/apiguardian`, `net/bytebuddy`), so supervised code may declare classes in namespaces Ares treats as infrastructure. Pre-existing, and neither introduced nor worsened here. Fixing it needs a boundary version bump with the Java list, both build fragments, tests and migration notes together, and it may break existing exercises, so it belongs in its own pull request. It should ship in the same release if possible.
- No path rule authenticates an unsigned artefact when the classpath is attacker-controlled. The supervised-origin check recognises the four conventional Maven and Gradle output layouts; a custom output directory or a project-owned JAR yields no supervised root. This narrows the partial-spoof surface and does not widen the existing one, but it is not an authentication mechanism.
- The resolved declaration is best-effort normalisation for ordinary public virtual calls on platform classes, not a reconstruction of dispatch. `super.method()`, bridge methods, private or package-private hiding, null receivers and constructors all fall back to today's behaviour.

## Checklist

## Review progress

- [ ] Code review
- [ ] Manual test

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>de.tum.cit.ase</groupId>
 	<artifactId>ares</artifactId>
-	<version>2.1.1</version>
+	<version>2.1.2</version>
 	<properties>
 		<!-- 0. General Project Settings -->
 		<maven.compiler.source>17</maven.compiler.source>
--- a/src/main/java/de/tum/cit/ase/ares/api/aop/java/aspectj/adviceandpointcut/JavaAspectJAbstractAdviceDefinitions.aj
+++ b/src/main/java/de/tum/cit/ase/ares/api/aop/java/aspectj/adviceandpointcut/JavaAspectJAbstractAdviceDefinitions.aj
@@ -674,22 +674,6 @@ public abstract aspect JavaAspectJAbstractAdviceDefinitions {
 	// <editor-fold desc="Signature handling (AspectJ-only)">
 
 	/**
-	 * Formats a join-point signature in the same shape produced by the Byte Buddy /
-	 * Instrumentation advice:
-	 * {@code <declaringType>.<methodName>(<paramType1>,<paramType2>,...)}.
-	 * <p>
-	 * AspectJ's {@link org.aspectj.lang.Signature#toLongString()} prepends Java
-	 * modifiers (e.g. {@code "public transient "}) and, for constructors, omits
-	 * {@code .<init>} altogether
-	 * ({@code "java.lang.ProcessBuilder(java.lang.String[])"}). The Instrumentation
-	 * side and the JSON expectation files use {@code <init>}-style signatures, so
-	 * this helper converts AspectJ's join-point shape to that form.
-	 *
-	 * @param sig the AspectJ {@link org.aspectj.lang.Signature} from the join point
-	 * @return normalized signature string with no leading modifiers and an explicit
-	 *         {@code <init>} marker for constructors
-	 */
-	/**
 	 * Appends the resolved runtime declaration of a denied call to its static
 	 * signature, when the two differ.
 	 * <p>
@@ -775,6 +759,22 @@ public abstract aspect JavaAspectJAbstractAdviceDefinitions {
 		}
 	}
 
+	/**
+	 * Formats a join-point signature in the same shape produced by the Byte Buddy /
+	 * Instrumentation advice:
+	 * {@code <declaringType>.<methodName>(<paramType1>,<paramType2>,...)}.
+	 * <p>
+	 * AspectJ's {@link org.aspectj.lang.Signature#toLongString()} prepends Java
+	 * modifiers (e.g. {@code "public transient "}) and, for constructors, omits
+	 * {@code .<init>} altogether
+	 * ({@code "java.lang.ProcessBuilder(java.lang.String[])"}). The Instrumentation
+	 * side and the JSON expectation files use {@code <init>}-style signatures, so
+	 * this helper converts AspectJ's join-point shape to that form.
+	 *
+	 * @param sig the AspectJ {@link org.aspectj.lang.Signature} from the join point
+	 * @return normalized signature string with no leading modifiers and an explicit
+	 *         {@code <init>} marker for constructors
+	 */
 	@Nonnull
 	protected static String formatSignature(@Nonnull Signature sig) {
 		return formatSignature(sig, sig.getDeclaringTypeName());
--- a/src/main/java/de/tum/cit/ase/ares/api/aop/java/aspectj/adviceandpointcut/JavaAspectJAbstractAdviceDefinitions.aj
+++ b/src/main/java/de/tum/cit/ase/ares/api/aop/java/aspectj/adviceandpointcut/JavaAspectJAbstractAdviceDefinitions.aj
@@ -15,8 +15,10 @@ import java.util.concurrent.ConcurrentHashMap;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.Signature;
 import org.aspectj.lang.reflect.CodeSignature;
+import org.aspectj.lang.reflect.MethodSignature;
 //</editor-fold>
 
 public abstract aspect JavaAspectJAbstractAdviceDefinitions {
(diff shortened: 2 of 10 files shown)


MarkusPaulsen:
@coderabbitai full review

MarkusPaulsen:
@coderabbitai Thanks. Two of the three are fixed in eb23cfd0; the third I am declining, with reasons.

**Fixed — fixture failures reported as skips.** Correct on both counts. `testCallgraphBuilderInitialisesWithoutOptionalJqwikOnTheClasspath` called `currentTestClassPath()` twice and compared the results, so a failure to *derive* the runner classpath surfaced as a confusing size mismatch rather than the missing fixture it is. It now captures once and assumes the derived classpath still contains the origins it intends to remove. `testOptionalFrameworkOriginsAreDiscardedWhenAnyClassIsAbsent` gained the matching assumption. jqwik being absent is precisely the consumer situation this PR fixes, so asserting it was the wrong shape.

**Fixed — booter JAR manifest.** Also correct, both halves. `JarFile.getManifest()` returns `null` for a JAR without `META-INF/MANIFEST.MF`, and a single `.jar` entry does not prove Surefire booter JAR, so that was a latent `NullPointerException`. Manifest `Class-Path` entries are relative URLs by specification, so `Path.of(URI)` would have thrown `IllegalArgumentException`; non-absolute and non-`file:` entries are now skipped.

**Declining — consolidating the denial-suffix concatenation.** The observation is accurate: the `+ (studentCalledMethod == null ? ...) + " | " + reason` pattern does repeat across roughly 14 sites. Three reasons not to do it here:

1. The duplication is pre-existing. Before this PR those sites read `fullMethodSignature + (studentCalledMethod == null ...)`. This PR replaced one token per site; it did not create the pattern.
2. These are the denial paths of the security enforcement layer. Restructuring all 14 in a PR that already carries two unrelated fixes enlarges a security-sensitive diff for zero behaviour change, and this PR is deliberately scoped so that reviewers can confirm the static signature never reaches an enforcement key.
3. The shape is not actually uniform, as the comment itself notes: two Command sites use a localised reason rather than `buildDenialReason(...)`, and `checkSinglePathRole` in the FileSystem advice orders its arguments differently. A single overload would not cover all of them cleanly.

Worth flagging: the committable suggestion in that comment does not compile. The `+` operators are missing from the concatenation, so it renders as `describeDeniedCall(joinPoint, fullMethodSignature)` followed by two bare expressions.

Happy to do the consolidation as a separate follow-up PR if a maintainer wants it.

MarkusPaulsen:
@coderabbitai review

Both of your actionable findings were addressed in eb23cfd0, and you accepted the reasoning for declining the third. The `CHANGES_REQUESTED` review from 11:26Z predates that commit, so please re-evaluate against the current head.

For context since that review:
- eb23cfd0 fixed the fixture assumptions and the booter-JAR manifest handling.
- CI is green on eb23cfd0 across all four mode combinations, plus CodeQL and the Coverage Report.
- Section 5 of the description now carries the real per-class figures from the Coverage Report artefact rather than a local measurement.

Yes please to the follow-up issue for consolidating the denial-suffix construction.

PR #175, opened 2026-08-04
Fix Phobos timeout unit serialization

ShudongCai:
## Summary

  Ares expresses policy timeouts in milliseconds, but wrote the raw millisecond integer into the generated Phobos configuration, which reads that field as seconds. The value is now serialised as canonical decimal seconds, so a 3000 ms policy limit produces `timeout=3.000` instead of `timeout=3000`.

  ## Linked issues

  None. This change is coordinated with the Phobos runtime update in https://github.com/ls1intum/phobos/pull/2, which must merge and ship before this behaviour is released.

  ## 1. Problem

  `ResourceLimitsPermission` is a millisecond value. The policy schema documents it that way (`- timeout: 10000  # REQUIRED (milliseconds)`, `SecurityPolicyManual.md` section 8.6), and the record rejects anything that is not strictly positive.

  `JavaPhobosTestCase.appendLimitsSection` previously emitted that number verbatim into the `[limits]` section of the generated `SpecificExercise.cfg`. The Phobos sandbox parses that field as **seconds** and hands it to GNU `timeout`. A policy that says `timeout: 3000` therefore described three seconds and
  produced a three thousand second limit, a factor of one thousand too permissive.

  Observed with `examples/ares-exercise-maven` (Java 21, Maven, ArchUnit + AspectJ, Windows), whose policy already carries `regardingTimeouts: - timeout: 3000`. Running the generation entry point on `main` writes `timeout=3000`; the expected content is `timeout=3.000`.

  The root cause sits in the **serialisation boundary between the policy layer and the generated Phobos configuration**, not in the enforcement layer. Neither the AOP nor the architecture path is involved, and the policy model itself was already correct.

  This is a **false negative**: the limit is too loose, so supervised code runs far longer than the instructor specified. It is not a false positive and cannot fail a correct submission.

  Scope of the live impact, stated precisely so the severity is not overread: in Ares 2.1.x the Phobos family is a **generation-only** stage. `JavaTestCaseFactoryAndBuilder.executeTestCases` dispatches only the architecture and AOP cases, so a normal `mvn test` run through the JUnit extension is not bounded
  by this value today (`SecurityPolicyManual.md` section 8.6 documents this). The defect is therefore latent in the in-process path, and live for any consumer that feeds the generated configuration to a real Phobos sandbox. It becomes a live weakening the moment Phobos dispatch is enabled, which is why it
  is worth fixing at the boundary now rather than after that migration.

  Instructors get the limit they wrote. A policy saying `timeout: 3000` yields a three second sandbox limit rather than a fifty minute one, and sub-second limits survive instead of truncating: `timeout: 500` becomes `0.500` rather than collapsing to `0`, which Phobos treats as "disabled".

  Students are unaffected in either direction. Nothing that passed before now fails, because the value only ever bounded execution in the Phobos sandbox path.

  The millisecond to second conversion now happens in exactly one place, `serialiseLimitValue`, with the cross-repository contract written down at that point. Previously the unit boundary was implicit and split across two repositories, so neither side documented who owned the conversion.

  The output is a canonical, locale-independent `S.mmm` form produced with `Locale.ROOT`, so it cannot drift on a machine whose default locale uses a decimal comma. The Phobos side validates exactly that shape, which makes an accidental double conversion a parse error rather than a silent thousandfold
  change.

  ## 4. Testing manual

  **Prerequisites**

  1. JDK 21 and Maven.
  2. This branch, `fix/i-090-timeout-units`, checked out in the Ares repository.
  3. `examples/ares-exercise-maven` as shipped. Its policy already contains `regardingTimeouts: - timeout: 3000`, so nothing has to be authored.
  4. No Docker, no Phobos runtime and no echo server are needed. This change is observed in a generated file.

  **Steps**

  1. Install the Ares build from this branch into the local repository, from the repository root:

     ```
     mvn -B -ntp install -DskipTests
     ```

     Expected result: `BUILD SUCCESS`. Note the project version (currently `2.1.2`); it is used as `<ARES_VERSION>` below.

  2. Copy the example out of the repository so the generated files do not dirty the working tree, and change into the copy:

     ```
     cp -r examples/ares-exercise-maven /path/to/scratch/genproj
     cd /path/to/scratch/genproj
     ```

  3. Resolve the example against the build from step 1:

     ```
     mvn -q -B -ntp -Dares.version=<ARES_VERSION> dependency:build-classpath -Dmdep.outputFile=cp.txt
     ```

     Expected result: `cp.txt` is written and the build succeeds.

  4. Run the generation entry point, which is what writes the Phobos configuration. Use `:` as the classpath separator on Linux and macOS, `;` on Windows:

     ```
     java -cp "target/classes:$(cat cp.txt)" de.tum.cit.ase.ares.api.Main src/test/resources/SecurityPolicy.yaml .
     ```

     Expected result: exit code 0. Two files appear in the project root, `SpecificExercise.cfg` and `PhobosCopyTool.sh`.

  5. Read the generated configuration:

     ```
     cat SpecificExercise.cfg
     ```

     Expected result: the `[limits]` section carries decimal seconds, not the raw millisecond integer:

     ```
     [readonly]
     allowed.txt

     [limits]
     timeout=3.000
     ```

     On `main` the same step yields `timeout=3000`. That single line is the whole change.

  6. Vary the policy value to confirm the conversion is exact rather than a whole-second rounding. Edit `regardingTimeouts` in `src/test/resources/SecurityPolicy.yaml`, delete `SpecificExercise.cfg`, and repeat step 4 for each value.

     Expected result, confirmed for all three:

     | Policy value | Generated line |
     |---|---|
     | `timeout: 2000` | `timeout=2.000` |
     | `timeout: 1234` | `timeout=1.234` |
     | `timeout: 500` | `timeout=0.500` |

  7. Run the focused unit test from the Ares repository root:

     ```
     mvn -o test -Dtest=JavaPhobosTestCaseTest
     ```

     Expected result: `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

  **Expected result**

  Stated per step above. The single observable claim is the `timeout=` line of the generated `SpecificExercise.cfg`, read directly from the file.

  1. **The conversion must happen once, not twice.** `timeout: 3000` must produce `timeout=3.000`, never `timeout=0.003`. A second division would show up immediately in step 5.
  2. **An absent limit must not invent one.** Set `regardingTimeouts: [ ]`, delete `SpecificExercise.cfg` and repeat step 4. Confirmed: the generated file contains **no `[limits]` section at all**, so Phobos keeps its existing disabled behaviour rather than receiving `timeout=0.000`.
  3. **Non-positive limits stay rejected at the API boundary.** `new ResourceLimitsPermission(0)` and `new ResourceLimitsPermission(-1)` must still throw `IllegalArgumentException`; `JavaPhobosTestCaseTest` asserts both. The serialisation change deliberately did not loosen that validation.
  4. **Sub-second limits must not silently become "no limit".** `timeout: 500` must produce `0.500`. Integer-second truncation would produce `0`, which Phobos reads as disabled, and that would be a security regression rather than a rounding detail.

  **Modes exercised**

  No mode-specific behaviour changed. `serialiseLimitValue` runs during Phobos configuration generation and is reached identically regardless of the architecture or AOP mode; neither an ArchUnit rule nor an AspectJ or instrumentation advice is involved. All four combinations were nevertheless run by CI on
  this branch and passed.

  

  Read from `site/jacoco/jacoco.csv` in the `coverage-report` artefact of the Coverage Report job on this branch (workflow run `30921639445`). `JavaPhobosTestCase` is the only production class this pull request changes; the nested `JavaPhobosTestCase.Builder` is reported separately by JaCoCo and is
  untouched here, so it is not listed.

  | Class | Instruction coverage | Branch coverage | Line coverage | Complexity coverage | Method coverage | Confirmation (meaningful assertions) |
  | --- | ---: | ---: | ---: | ---: | ---: | :---: |
  | `de.tum.cit.ase.ares.api.phobos.JavaPhobosTestCase` | 91.9% (399/434) | 85.0% (17/20) | 95.2% (79/83) | 89.3% (25/28) | 100.0% (17/17) | yes |

  Both branches of `serialiseLimitValue` are taken: the `timeout` key by the three serialisation assertions, and the non-`timeout` key by the existing file-system and network cases that emit their values verbatim. The assertions compare the exact generated string (`2.000`, `1.234`, `0.500`) rather than
  asserting mere execution, and the sub-second case specifically pins the value that integer truncation would destroy.

  **Yes, for the generated Phobos configuration.** This is not a source-compatibility break:

  - the public API under `de.tum.cit.ase.ares.api` is unchanged (`serialiseLimitValue` is private, and `ResourceLimitsPermission` keeps its millisecond contract);
  - the security policy file format and schema are unchanged, and instructors keep writing milliseconds exactly as documented;
  - the minimum JDK, Maven and Gradle versions are unchanged.

  What changes is the **content of the generated `SpecificExercise.cfg`**. A Phobos runtime that predates https://github.com/ls1intum/phobos/pull/2 accepts only integer seconds and would not parse `3.000`, so the two must be released in order: **the updated Phobos runtime image first, this change second**.
  Instructors do not have to touch an existing exercise; no policy file needs editing.

  Existing native Phobos configurations that use integer seconds remain valid, and integer values are not reinterpreted as milliseconds, so hand-written Phobos configuration is unaffected.

  ## Known follow-up, not addressed here

  Ares bundles a snapshot of the Phobos runtime scripts under `src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/`. Those copies still carry the old integer-only parser (`phobos-common.sh`, the `^timeout[[:space:]]*=[[:space:]]*([0-9]+)$` branch) and still invoke GNU `timeout` without the explicit
  seconds suffix (`phobos-filesystem.sh`).

  They are **not deployed**: `PhobosCopyFiles.csv` copies only `PhobosCopyTool.sh` and `PhobosEditFiles.csv` generates only `SpecificExercise.cfg`, which the run in section 4 confirms, since those two files are the only ones written. They are exercised solely by `PhobosShellContractTest`, which asserts
  unknown-section rejection and fail-closed runtime errors and does not touch timeout parsing, so this pull request neither breaks nor is broken by them.

  They will nonetheless drift from `ls1intum/phobos` once the coordinated pull request merges. Resynchronising that snapshot belongs in its own change, together with a decision on whether Ares should vendor those scripts at all.

  ## Checklist

  - [x] Documentation (`docs/`, `README.adoc`, Javadoc) was updated where the change is user-facing. Javadoc on `serialiseLimitValue` records the cross-repository contract. The policy documentation already states milliseconds and stays correct; the generated configuration format is not described in `docs/`.
  - [x] CI is green, or every remaining failure is explained above. All fifteen checks pass, including all four mode combinations, both example exercises and the Coverage Report job.

  ## Review progress

  - [ ] Code review
  - [ ] Manual test

### DIFF
--- a/src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
+++ b/src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
@@ -240,8 +240,35 @@ public class JavaPhobosTestCase extends PhobosTestCase {
 			return;
 		}
 		out.append('[').append(header).append("]\n");
-		limits.forEach((k, v) -> out.append(k).append('=').append(v).append('\n'));
+		limits.forEach((k, v) -> out.append(k).append('=').append(serialiseLimitValue(k, v)).append('\n'));
 		out.append('\n');
+	}
+
+	/**
+	 * Serialises a single {@code [limits]} value for the Phobos configuration.
+	 * <p>
+	 * The {@code timeout} limit is defined in <em>milliseconds</em> by
+	 * {@link ResourceLimitsPermission}, whereas the Phobos configuration field is
+	 * expressed in <em>seconds</em>. This method is the single authoritative
+	 * conversion point: milliseconds are rendered as canonical decimal seconds
+	 * {@code S.mmm} (exactly three fractional digits, formatted with
+	 * {@link java.util.Locale#ROOT} so it never depends on the platform locale, and
+	 * containing only digits and one dot). The Phobos runtime validates that exact
+	 * representation and appends an {@code s} suffix before handing it to GNU
+	 * {@code timeout}, so no second conversion happens anywhere else and no
+	 * sub-second precision is lost. All other keys are emitted verbatim.
+	 *
+	 * @param key   the limit key, e.g. {@code "timeout"}
+	 * @param value the limit value; for {@code timeout} this is milliseconds
+	 * @return the serialised value written after {@code key=}
+	 */
+	private static String serialiseLimitValue(String key, long value) {
+		if (!"timeout".equals(key)) {
+			return Long.toString(value);
+		}
+		long seconds = value / 1000L;
+		long milliseconds = value % 1000L;
+		return String.format(java.util.Locale.ROOT, "%d.%03d", seconds, milliseconds);
 	}
 
 	/**
--- a/src/test/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCaseTest.java
+++ b/src/test/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCaseTest.java
@@ -1,9 +1,13 @@
 package de.tum.cit.ase.ares.api.phobos;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.junit.jupiter.api.Test;
 
@@ -34,7 +38,45 @@ class JavaPhobosTestCaseTest {
 		JavaPhobosTestCase timeout = JavaPhobosTestCase.builder()
 				.javaPhobosTestCaseSupported(JavaPhobosTestCaseSupported.TIMEOUT)
 				.resourceAccessSupplier(() -> List.of(new ResourceLimitsPermission(1234))).build();
-		assertTrue(timeout.writePhobosTestCase().contains("timeout=1234"));
+		assertTrue(timeout.writePhobosTestCase().contains("timeout=1.234"));
+	}
+
+	@Test
+	void timeoutIsSerialisedAsCanonicalDecimalSeconds() {
+		// ResourceLimitsPermission carries milliseconds, but the Phobos [limits]
+		// timeout
+		// field is seconds. The generator converts once at this serialisation boundary,
+		// emitting canonical S.mmm decimal seconds (locale-independent, full
+		// millisecond
+		// precision, digits and one dot only).
+		assertEquals("2.000", timeoutValue(2000));
+		assertEquals("10.000", timeoutValue(10000));
+		assertEquals("2.500", timeoutValue(2500));
+		// Sub-second must not collapse to zero (which Phobos would treat as disabled).
+		assertEquals("0.500", timeoutValue(500));
+		assertEquals("0.001", timeoutValue(1));
+
+		// An absent timeout yields no [limits] section; its default behaviour is
+		// decided
+		// elsewhere, not at this serialisation boundary.
+		JavaPhobosTestCase none = JavaPhobosTestCase.builder()
+				.javaPhobosTestCaseSupported(JavaPhobosTestCaseSupported.TIMEOUT)
+				.resourceAccessSupplier(() -> List.of()).build();
+		assertFalse(none.writePhobosTestCase().contains("[limits]"));
+
+		// Non-positive milliseconds are rejected at the API boundary (unchanged).
+		assertThrows(IllegalArgumentException.class, () -> new ResourceLimitsPermission(0));
+		assertThrows(IllegalArgumentException.class, () -> new ResourceLimitsPermission(-1));
+	}
+
+	private static String timeoutValue(long millis) {
+		JavaPhobosTestCase timeout = JavaPhobosTestCase.builder()
+				.javaPhobosTestCaseSupported(JavaPhobosTestCaseSupported.TIMEOUT)
+				.resourceAccessSupplier(() -> List.of(new ResourceLimitsPermission(millis))).build();
+		String cfg = timeout.writePhobosTestCase();
+		Matcher matcher = Pattern.compile("(?m)^timeout=(\\S+)$").matcher(cfg);
+		assertTrue(matcher.find(), "generated cfg must contain a timeout line, was:\n" + cfg);
+		return matcher.group(1);
 	}
 
 	@Test


src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh
```diff
@@ -31,20 +31,96 @@ limits
...
+# grammar; anything else is section content, which is what keeps a path that
+# merely contains brackets a path. A header must be exactly one bracket pair
+# around a name that is not empty once trimmed.
+#
+# Prints the normalised name and returns 0 for a header, 1 for ordinary content,
+# 2 for a header-looking line that is malformed, and 3 for one that names
+# nothing.
+classify_config_line() {
+  local line="$1" name
+  [[ "$line" == '['* ]] || return 1
```
Claudia-Anthropica, line 67: @ShudongCai [medium] This classifies every value beginning with `[` as a malformed header. `FILE_PATH_PATTERN` permits relative paths such as `[draft`, and `JavaPhobosTestCase` emits them verbatim under `[readonly]` or `[write]`, so a valid policy now exits with `PHB-EPOLICY` instead of sandboxing that path—a breaking false positive.

In `src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh`, `classify_config_line` treats every bracket-leading line as a header, so valid generated `FilePermission` paths such as `[draft` are rejected. Prefix relative bracket-leading paths with `./` when `JavaPhobosTestCase` emits them and add a generator-to-parser regression test.

</details>

src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh
```diff
@@ -53,10 +129,21 @@ parse_cfg_policy() {
...
   : >"$ro"; : >"$rw"; : >"$hide"; : >"$net"
   local sec=""
-  while IFS= read -r line || [[ -n "$line" ]]; do
-    line="${line%%#*}"; line="$(echo "$line" | sed -E 's/^[[:space:]]+//; s/[[:space:]]+$//')"
+  local raw line name status
+  # Section-local, so it never carries a count from another section or file.
+  local timeout_values=0
+  # Per-file, so the caller can tell "this policy set no timeout" from "this
+  # policy set a timeout of zero".
+  PARSED_TIMEOUT_SET=0
```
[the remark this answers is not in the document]
ShudongCai: Thanks for the suggestion. I verified this with ShellCheck 0.11.0.

`phobos-common.sh` is a sourced fragment, and `PARSED_TIMEOUT_SET` is consumed by `phobos.sh` when layered policies are combined. Analysing the fragment alone reports SC2034 for ten cross-file variables; the version on `main` already reports nine of the same warnings. By contrast, `shellcheck -x phobos.sh` follows the existing source directive and reports no SC2034 for `PARSED_TIMEOUT_SET`.

ShellCheck is also not configured as a project CI gate. Adding a suppression for this single assignment would therefore be inconsistent with the other cross-file variables and would not make the standalone fragment pass ShellCheck.

No change is needed here; the correct ShellCheck entry point is the sourcing script with `-x`.

src/test/java/de/tum/cit/ase/ares/api/phobos/PhobosShellContractTest.java
```diff
@@ -47,6 +51,517 @@ void missingBaseAndRuntimeDependenciesUseDocumentedFailClosedErrors() throws Exc
...
+				slowCommandTouchingMarker(shellRoot));
+
+		assertEquals(0, longer.exitCode(), longer.output());
+		assertTrue(Files.exists(markerOf(shellRoot)), "the later, longer timeout must replace the base one");
+
+		ProcessResult shorter = runWrapper(shellRoot, "shorter.cfg", "[timeout]\n0.200\n", "sleep 5");
+
+		assertEquals(14, shorter.exitCode(), shorter.output());
+		assertTrue(shorter.output().contains("Timed out after 0.200s"), shorter.output());
+	}
```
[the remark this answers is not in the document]
ShudongCai: I verified the missing-tool path, but GNU timeout is a required conditional runtime dependency of the shipped wrapper, not an optional external fixture.

When it is unavailable, the wrapper already distinguishes the failure explicitly: it exits with code 15, reports PHB-ERUNTIME, and does not start the protected command. A real enforced timeout instead exits with code 14 and reports PHB-ETIMEOUT. The assertions pass the complete process output as their failure message, so a missing executable is not reported as an enforcement defect.

The cited path instruction concerns optional external network fixtures, such as the echo server. It does not apply to required local runtime executables. Skipping here would hide a broken supported environment. Moreover, command -v timeout would not establish GNU compatibility with the --kill-after option used by the wrapper.

The existing hard assertions therefore remain unchanged.

src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh
Claudia-Anthropica, line 67: @ShudongCai [medium] This intentionally reserves every bracket-leading line as a header, but native `Base*.cfg` and `--config` files bypass `JavaPhobosTestCase.serialiseFilesystemPath`. A previously accepted filesystem entry such as `[read]\n[draft` now exits with `PHB-EPOLICY`; the PR instead claims hand-written Phobos configurations are unaffected and provides no migration guidance for spelling the path as `./[draft`.

In `src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh`, the stricter header grammar introduces an undocumented breaking change for hand-written configurations containing bracket-leading filesystem paths. Document the `./` escaping requirement in the user-facing Phobos configuration documentation and add this native-configuration migration to the PR's breaking-changes section.

</details>

Claudia-Anthropica: @ShudongCai Thanks, the generator now prefixes bracket-leading paths and the generator-to-parser regression covers the original policy-model issue.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
```diff
@@ -139,10 +139,33 @@ private static void appendFilesystemSection(StringBuilder out, String header, Se
...
+	 * necessarily relative, so no host-dependent absolute-path check is involved:
+	 * generation may run on Windows while the configuration targets Linux. Every
+	 * other path, absolute or relative, is emitted verbatim, including one that
+	 * merely contains a bracket later on.
+	 *
+	 * @param path the path collected from the permission model
+	 * @return the path as written into a filesystem section
+	 */
+	private static String serialiseFilesystemPath(String path) {
+		return path.startsWith("[") ? "./" + path : path;
```
Claudia-Anthropica: @ShudongCai [medium] `FILE_PATH_PATTERN` accepts paths with leading whitespace, but this checks only the raw first character. A valid path such as `" [draft"` is emitted unchanged; `preprocess_config_line` strips the leading space, and `classify_config_line` then rejects `[draft` as a malformed header, causing a false-positive policy failure. Prefix relative paths starting with parser-trimmed whitespace as well, so `./ [draft` preserves the actual filename, and cover this generator-to-parser case.

In `src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java`, `serialiseFilesystemPath` leaves valid whitespace-leading paths unchanged, allowing parser trimming to change or reject them. Prefix raw relative paths beginning with whitespace as well as `[` using `./`, and add a generator-to-parser regression for `" [draft"`.

</details>

src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh
Claudia-Anthropica: @ShudongCai The Overview now documents the escape, but the current PR breaking-changes section still says hand-written Phobos configuration is unaffected, so this remains unresolved.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
```diff
@@ -139,10 +139,42 @@ private static void appendFilesystemSection(StringBuilder out, String header, Se
...
+	 *
+	 * @param path the path collected from the permission model
+	 * @return the path as written into a filesystem section
+	 */
+	private static String serialiseFilesystemPath(String path) {
+		if (path.isEmpty()) {
+			return path;
+		}
+		char first = path.charAt(0);
+		return first == '[' || first == ' ' || first == '\t' ? "./" + path : path;
```
Claudia-Anthropica: @ShudongCai [medium] `preprocess_config_line` strips every leading POSIX `[[:space:]]` character, but this helper protects only ASCII space and tab. `FILE_PATH_PATTERN` also accepts paths beginning with carriage return (`\r`), vertical tab (`\u000B`), or form feed (`\f`); these are emitted unchanged and then stripped, so a permission such as `\u000Bsecret` targets `secret` instead, while `\u000B[draft` becomes a malformed header. This can grant the wrong path or reject a valid policy.

In `src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java`, `serialiseFilesystemPath` does not protect leading carriage return, vertical tab, or form feed even though the shell parser strips them. Extend the prefix predicate to cover `\r`, `\u000B`, and `\f`, and add generator-to-parser regression tests proving each character survives unchanged.

</details>

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
```diff
@@ -139,10 +139,48 @@ private static void appendFilesystemSection(StringBuilder out, String header, Se
...
+	 *
+	 * @param path the path collected from the permission model
+	 * @return the path as written into a filesystem section
+	 */
+	private static String serialiseFilesystemPath(String path) {
+		if (path.isEmpty()) {
+			return path;
+		}
+		char first = path.charAt(0);
+		boolean reserved = first == '[' || first == ' ' || first == '\t' || first == '\r' || first == '\u000B'
```
Claudia-Anthropica, line 179: @ShudongCai [medium] This list is not equivalent to the parser's `[[:space:]]` trimming because that character class is locale-sensitive. With `C.UTF-8`, Bash also trims U+2003 EM SPACE; `FILE_PATH_PATTERN` accepts `\u2003secret`, this helper emits it unchanged, and parsing silently retargets the permission to `secret`, risking an unintended grant. Make the shell trimming locale-invariant and limited to the same explicit six characters, then cover a UTF-8 Unicode-space path in the generator-to-parser contract test.

In `src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java`, `serialiseFilesystemPath` protects only six characters while the shell's locale-sensitive `[[:space:]]` may strip additional Unicode whitespace, causing accepted paths to be retargeted. Replace the parser's locale-sensitive trimming with trimming of the exact six documented characters and add a C.UTF-8 generator-to-parser regression.

</details>

src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/phobos-common.sh
Claudia-Anthropica: @ShudongCai Thanks, the breaking-changes section now documents the `./` migration for native bracket- and blank-leading paths.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
Claudia-Anthropica: @ShudongCai Thanks, `trim_ends` now uses the C locale and the UTF-8 regression confirms Unicode-leading paths remain unchanged.

Claudia-Anthropica (dismissed):
@ShudongCai The exact-head follow-up handles the unavailable UTF-8 locale fixture without masking parser failures, and the documentation correction is sound. The timeout conversion, fail-closed parser behaviour, path serialisation, and associated regressions remain consistent with the resolved review findings. The snapshot shows successful Build and Maven example checks, with no required checks or completed relevant failures.

MarkusPaulsen:
count towards it.

This description was written before those limits and exceeds 4 of them, so I have

<details><summary>Summary, as it read before (631 characters, limit 500)</summary>

Ares expresses policy timeouts in milliseconds but wrote the raw millisecond integer into the generated Phobos configuration, which reads that field as seconds. This pull request fixes that unit boundary and, in the course of proving it end to end, hardens the bundled Phobos policy parser so that unreadable, missing or ambiguous policy input fails closed instead of silently disabling a limit, dropping a section or renaming a path.

Four things change: the Java timeout serialisation boundary, the Java filesystem-path serialisation boundary, the bundled Phobos shell parser, and the test coverage that ties the halves together.

</details>

<details><summary>1. Problem, as it read before (2332 characters, limit 1000)</summary>

`ResourceLimitsPermission` is a millisecond value. The policy schema documents it that way (`- timeout: 10000  # REQUIRED (milliseconds)`, `SecurityPolicyManual.md` section 8.6), and the record rejects anything that is not strictly positive.

`JavaPhobosTestCase.appendLimitsSection` previously emitted that number verbatim into the `[limits]` section of the generated `SpecificExercise.cfg`. The Phobos sandbox parses that field as **seconds** and hands it to GNU `timeout`. A policy saying `timeout: 3000` therefore described three seconds and produced a three thousand second limit, a factor of one thousand too permissive.

Investigating that boundary surfaced further fail-open behaviour:

- an unrecognised timeout value was ignored, disabling the limit entirely instead of rejecting the policy;
- a dedicated `[timeout]` section holding no value, or only comments, parsed successfully and could silently retain a value from a previously parsed policy file;
- the parser could not distinguish "this policy set no timeout" from "this policy set a timeout of zero", so an explicit `0.000` in a later layer did not clear an earlier limit;
- validation and parsing disagreed about what a section header is, so a header could pass validation under one name and then be parsed as a different section, or as none at all, and its whole body was discarded in silence;
- a filesystem path whose first character is `[` or a blank was emitted verbatim, so the parser either read it as a section header or trimmed it into a different filename;
- the parser trimmed line ends with a locale-sensitive character class, so under a UTF-8 locale a path beginning with a Unicode separator such as U+2003 was silently renamed.

This is a **false-negative** class of defect throughout: limits too loose, or sections and paths silently absent or retargeted. None of it can fail a correct submission.

Scope of live impact, stated so the severity is not overread: in Ares 2.1.x the Phobos family is a **generation-only** stage. `JavaTestCaseFactoryAndBuilder.executeTestCases` dispatches only the architecture and AOP cases, so a normal `mvn test` run through the JUnit extension is not bounded by this value today. The defect is latent in the in-process path and live for any consumer that feeds the generated configuration to a real Phobos sandbox.

</details>

<details><summary>3. Improvement from the maintainer's perspective, as it read before (1025 characters, limit 1000)</summary>

The millisecond to second conversion happens in exactly one place, `serialiseLimitValue`, with the cross-repository contract written down at that point. The output is a canonical, locale-independent `S.mmm` form produced with `Locale.ROOT`, so it cannot drift on a machine whose default locale uses a decimal comma.

Filesystem paths are serialised through one helper, `serialiseFilesystemPath`, so the `readonly` and `write` sections are normalised identically. Its protected first-character set is written out explicitly rather than inferred from a general whitespace test, so it cannot rewrite a path the parser would have kept verbatim.

In the shell parser, section headers are recognised by a single shared grammar used by both validation and parsing, so the two can no longer disagree; the timeout cardinality rule and the timeout presence flag are each expressed once; and line trimming runs under a function-local `LC_ALL=C`, so the class of characters it removes is fixed rather than inherited from the environment.

</details>

<details><summary>Breaking changes and migration, as it read before (2641 characters, limit 1000)</summary>

This is not a source-compatibility break. The public API under `de.tum.cit.ase.ares.api` is unchanged, the security policy file format and schema are unchanged, instructors keep writing milliseconds exactly as documented, and the minimum JDK, Maven and Gradle versions are unchanged.

There are two runtime compatibility changes.

**1. Generated timeout values now use canonical decimal seconds.** What changes is the content of the generated `SpecificExercise.cfg`. A Phobos runtime that predates https://github.com/ls1intum/phobos/pull/2 accepts only integer seconds and would not parse `3.000`, so the two must be released in order: **the updated Phobos runtime image first, this change second**. Instructors do not have to touch an existing exercise; no policy file needs editing.

**2. Native filesystem entries beginning with a reserved or trimmed character need an explicit `./`.** The hardened parser reserves a line beginning with `[` for a section header and trims leading blank characters before reading a line. A relative filesystem entry whose first character is `[`, a space, a tab, a carriage return, a vertical tab or a form feed must therefore be written as `./[draft`, `./ [draft` and so on, in the `readonly`, `read`, `write`, `hide` and `tmpfs` sections of a hand-written `Base*.cfg` or a file passed through `--config`. Absolute paths, and entries whose brackets or blanks appear later such as `relative/[draft`, need no change. **Ares-generated configurations apply this serialisation automatically**, so this migration applies to hand-written configurations. It is documented in `docs/Overview.md`. Characters outside that set, such as the Unicode separator U+2003, are ordinary filename characters and are preserved rather than escaped, which is what the locale-invariant trimming guarantees.

**Where each artefact lives.** Ares generates and deploys `SpecificExercise.cfg` only: `PhobosCopyFiles.csv` copies `PhobosCopyTool.sh`, `PhobosEditFiles.csv` generates `SpecificExercise.cfg`, and `PhobosCopyTool.sh` copies that one file into the runtime directory. The scripts under `src/main/resources/de/tum/cit/ase/ares/api/templates/phobos/` are a bundled snapshot exercised by `PhobosShellContractTest`, and this pull request hardens that snapshot. The parser and `Base*.cfg` used at runtime come from the image built from the independent `ls1intum/phobos` repository, **which is not modified here**. That runtime still carries the older, permissive grammar and locale-sensitive trimming, so the coordinated release needs equivalent behaviour on that side before migration point 2 takes effect for a deployed sandbox.

</details>

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
MarkusPaulsen: Fixed at the current head. `JavaPhobosTestCase.serialiseFilesystemPath` no longer tests only for `[`: it prefixes `./` when the raw first character is any of `[`, space, tab, carriage return, vertical tab or form feed, so `" [draft"` is emitted as `"./ [draft"` and survives the parser trim. The generator-to-parser regressions in `PhobosShellContractTest` cover the space and tab cases.

The branch has just been merged with `main` to clear a conflict in `PhobosShellContractTest.java`; the fix above is unchanged by that merge.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
MarkusPaulsen: Fixed at the current head. The protected set is now written out explicitly as exactly `[`, space, tab, carriage return, vertical tab and form feed, so `\r`, `\u000B` and `\f` are prefixed rather than emitted unchanged. It is deliberately an explicit list rather than a general whitespace test, and the parser side trims under the C locale so the two ends agree on the same six characters.

The branch has just been merged with `main` to clear a conflict in `PhobosShellContractTest.java`; the fix above is unchanged by that merge.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
```diff
@@ -139,10 +139,48 @@ private static void appendFilesystemSection(StringBuilder out, String header, Se
...
+				|| first == '\f';
+		return reserved ? "./" + path : path;
```
Claudia-Anthropica: @ShudongCai [medium] `FILE_PATH_PATTERN` accepts embedded newline characters, but this helper emits them verbatim. A read-only path such as `safe\n[timeout]\n0` therefore becomes three configuration lines; the parser accepts the injected section and disables the effective timeout, allowing one permission value to silently alter another security domain. Reject LF-containing paths at the policy-validation boundary before generation and add a generator-to-parser regression proving they fail closed.

In `src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java`, `serialiseFilesystemPath` emits embedded LF characters accepted by the policy model as configuration record delimiters, allowing a file path to inject a timeout override. Reject LF-containing file paths during policy validation and add a generator-to-parser test proving such input cannot clear the timeout.

</details>

Claudia-Anthropica (changes requested):
@ShudongCai The previously discussed timeout, locale, and native-path migration issues are resolved, and all captured CI checks succeeded at this head. One remaining serializer gap lets a valid file path containing a newline inject additional Phobos configuration records, including a timeout override, so generation must fail closed.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
MarkusPaulsen: The mechanism is confirmed, but the stated impact is not, and the real one is worse.

`FILE_PATH_PATTERN` does accept LF: it is compiled with `DOTALL` and its character class `[^*\x00]` excludes only `*` and NUL. `serialiseFilesystemPath` then emits the value verbatim, because it inspects the first character alone.

The timeout route does not work, though. `writeConfiguration` emits `readonly`, `write`, `network`, then `limits`, so `[limits]` is always last and a later `timeout=` overrides an injected `[timeout]`. I checked this against the bundled parser rather than reasoning about it: with `[readonly]\nsafe\n[timeout]\n0` followed by `[limits]\ntimeout=5.000`, `PARSED_TIMEOUT` comes back `5.000`, identical to the run without the injection.

What does work is injecting a filesystem section, which nothing later overrides. A single read-only value of `safe\n[write]\n/etc` parses to read-only `safe` and **write `/etc`**, so a value in one permission list grants a permission in another that the policy never granted.

One limit on severity worth stating: these values come from `security-policy.yaml`, which is instructor configuration. `SECURITY.md` puts an adversary who controls the build and its configuration outside the threat model, so this is not a student-reachable bypass. It is a fail-open in a component whose whole design is to fail closed, and a policy that quietly grants more than it says is worth refusing outright.

On the fix: rejecting LF at the validation boundary is right, and `\r` should stay accepted, since `read` splits records on LF alone and a leading `\r` is already handled by the `./` prefix. Excluding LF from the character class is not sufficient on its own, because without `MULTILINE` the trailing `$` still matches before a final line terminator, so `"safe\n"` would continue to pass; the anchor needs to be `\z`. I am raising with the maintainer whether this belongs in this pull request, which is about timeout units, or in its own.

Claudia-Anthropica (changes requested):
@ShudongCai The timeout conversion, parser hardening, and earlier path-migration issues are resolved, and every captured CI check succeeded at this exact head. [medium] The latest reply correctly shows that the later generated `[limits]` section defeats the timeout-injection example, but `FILE_PATH_PATTERN` still accepts LF and `serialiseFilesystemPath` emits it verbatim, allowing a read-only value to inject a `[write]` entry for a base-writable path. Reject LF during policy validation with an absolute `\z` end anchor and add a filesystem cross-section regression before merging.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
Claudia-Anthropica: @MarkusPaulsen Thanks—the serialization order confirms that `[limits]` overwrites the injected timeout, so that part of my earlier impact statement was incorrect. The current head still accepts LF and emits it verbatim, after which the parser classifies an injected `[write]` section; targeting a path writable in the base can therefore broaden the exercise policy. I’m retaining the finding for that filesystem cross-section path: reject LF during policy validation with a `\z` anchor and add the corresponding generator-to-parser regression.

MarkusPaulsen: Fixed in https://github.com/ls1intum/Ares2/pull/261, which is based on main rather than this branch, since the defect reproduces without the timeout-unit work and the fix touches neither timeouts nor `serialiseFilesystemPath`.

The guard went to the serialisation boundary rather than to policy validation, for a reason worth recording here. `SecurityPolicyYAMLReader` expands `${PROJECT_ROOT}` inside `readYamlTree` before `SecurityPolicySchemaValidator.validate` runs, so a validator-level rule would refuse a project merely for being checked out into a directory whose name holds a line feed. That is legal, and `SecurityPolicyStrictSchemaTest.expandsProjectRootAsScalarDataAfterParsing` already exercises exactly such a root. `FILE_PATH_PATTERN` is also shared by every enforcement mode, and only the Phobos configuration is line-based. Guarding where the file is written refuses a path from the policy and a path from an expanded root alike, without imposing one format's limit on the others.

`appendFilesystemSection` now refuses a path holding a line feed and fails closed with a localised message naming the section and the path. Both new tests are in `JavaPhobosTestCaseTest`, and the refusing one fails against the unguarded generator, so it is a regression test rather than a restatement.

Claudia-Anthropica (approved):
@ShudongCai The exact-head changes correctly convert millisecond limits to decimal seconds, fail closed on malformed timeout input, and preserve timeout layering and filesystem-path semantics. The LF cross-section issue is real but already exists at the captured base and is neither introduced nor worsened here, so tracking its serialization-boundary fix separately is appropriate. The captured CI snapshot contains no failed or unfinished checks.

src/main/java/de/tum/cit/ase/ares/api/phobos/JavaPhobosTestCase.java
Claudia-Anthropica: @MarkusPaulsen Thanks—the captured base confirms that LF-delimited cross-section injection predates this PR, and the changes here neither introduce nor worsen it. Tracking the serialization-boundary fix in #261 is appropriate, so this thread can be resolved.

